PR #2, opened 2017-11-09
when verifying a method with 'exactly = 0', return verification match…

yannickvg:
… if a method is never called instead of an error that the method is never called

Thanks for creating Mockk. While using your library I wanted to verify that a certain method on the mock is never called by using verify(exactly = 0). This always resulted in a error "Verification failed: call 1 of 1. No calls for mockk... " while that should result in a verification success. Therefor I made a fix to check the min and max in case of no calls to the mock method. If they are 0, it returns a verification success instead of failure. When min and max are not 0, everything stays the same and a failure is reported. I also added a unit test case for this fix.

### DIFF
--- a/mockk/src/main/kotlin/io/mockk/impl/Verification.kt
+++ b/mockk/src/main/kotlin/io/mockk/impl/Verification.kt
@@ -92,11 +92,15 @@ internal class UnorderedCallVerifierImpl(private val gw: MockKGatewayImpl) : Cal
         }
         val result = when (allCallsForMockMethod.size) {
             0 -> {
-                if (allCallsForMock.isEmpty()) {
-                    VerificationResult(false, "$callIdxMsg No calls for $mock/${call.matcher.method.toStr()}")
+                if (min == 0 && max == 0) {
+                    VerificationResult(true)
                 } else {
-                    VerificationResult(false, "$callIdxMsg No calls for $mock/${call.matcher.method.toStr()}.\n" +
-                            "Calls to same mock:\n" + formatCalls(allCallsForMock))
+                    if (allCallsForMock.isEmpty()) {
+                        VerificationResult(false, "$callIdxMsg No calls for $mock/${call.matcher.method.toStr()}")
+                    } else {
+                        VerificationResult(false, "$callIdxMsg No calls for $mock/${call.matcher.method.toStr()}.\n" +
+                                "Calls to same mock:\n" + formatCalls(allCallsForMock))
+                    }
                 }
             }
             1 -> {
--- a/mockk/src/test/kotlin/io/mockk/MockKTestSuite.kt
+++ b/mockk/src/test/kotlin/io/mockk/MockKTestSuite.kt
@@ -174,6 +174,9 @@ class MockKTestSuite : StringSpec({
         verify(exactly = 0, inverse = true) {
             mock.otherOp(0, 2)
         }
+        verify(exactly = 0) {
+            mock.neverCalled()
+        }
     }.config(enabled = true)
 
     "stubbing actions" {
@@ -659,4 +662,6 @@ class MockCls {
     fun chainOp(a: Int = 1, b: Int = 2) = if (a + b > 0) MockCls() else MockCls()
     fun arrayOp(array: Array<Any>): Array<Any> = array.map { (it as Int) + 1 }.toTypedArray()
     fun arrayOp(array: Array<Array<Any>>): Array<Array<Any>> = array.map { it.map { ((it as Int) + 1) as Any }.toTypedArray() }.toTypedArray()
+
+    fun neverCalled():Int = 1
 }


mockk/src/main/kotlin/io/mockk/impl/Verification.kt
```diff
@@ -92,11 +92,15 @@ internal class UnorderedCallVerifierImpl(private val gw: MockKGatewayImpl) : Cal
...
         val result = when (allCallsForMockMethod.size) {
             0 -> {
-                if (allCallsForMock.isEmpty()) {
-                    VerificationResult(false, "$callIdxMsg No calls for $mock/${call.matcher.method.toStr()}")
+                if (min == 0 && max == 0) {
+                    VerificationResult(true)
                 } else {
-                    VerificationResult(false, "$callIdxMsg No calls for $mock/${call.matcher.method.toStr()}.\n" +
-                            "Calls to same mock:\n" + formatCalls(allCallsForMock))
+                    if (allCallsForMock.isEmpty()) {
```
oleksiyp: Great. I was wondering what can be the use case when was writing that, but skipped thinking a lot. 

Can you just put all the ifs on one level?

oleksiyp:
@yannickvg I've merged PR, will fix ifs myself and do release soon

oleksiyp:
@yannickvg version 1.4 with your fix is released

yannickvg:
Thanks!

oleksiyp:
Hi @yannickvg 

btw what was your overall impression from trying the library out?
Was there too much difficulties? Was some positive impressions?
I'd like to make it better and that's impossible without side view. Mine is too biased.

Thanks,
Oleksii

yannickvg:
Hi Oleksii,

overall impression was really good. I didn’t write any complicated test
yet, but for the basic tests i wrote until now, it was working fine!

I’ll keep you updated about any remarks while I use it for further testing!

cheers,

Yannick

On Thu, Nov 16, 2017 at 8:55 AM, Oleksiy Pylypenko <notifications@github.com

yannickvg:
Hi Oleksii,

just ran into a feature that could be useful: spies

In Mockito you can create a spy, which is a hybrid between a mock and the
real object, so it has the real implementations and it tracks interactions
with it, but you can choose to mock certain functions on it.

I suppose its cleaner to use mocks, but i’ve noticed that it is sometimes
handy to mock certain functions of a class.

you think that would be useful?

cheers,

Yannick

oleksiyp:
Try `val spy = spyk(YourClass())` :-)

oleksiyp:
Seems docs really missing a lot...

yannickvg:
Awesome that it exists!

is there a way to verify that no interactions have been done on a mock?
(not specifying a method)



On Thu, Nov 23, 2017 at 3:56 PM, Oleksiy Pylypenko <notifications@github.com

yannickvg:
I found this in the docs

verify {
    listOf(obj2, obj3) wasNot Called}


but it doesn’t seem to compile

oleksiyp:
It was added this morning. Try 1.5.4

23 нояб. 2017 г. 4:29 PM пользователь "Yannick Van Godtsenhoven" <
notifications@github.com> написал:

> I found this in the docs
>
> verify {
> listOf(obj2, obj3) wasNot Called}
>
>
> but it doesn’t seem to compile
>
>
> On Thu, Nov 23, 2017 at 4:26 PM, Yannick Van Godtsenhoven <

yannickvg:
cool, really like the wasNot Called. nice readability of the tests

was trying out the spyk() but with less success. I also use dagger, so in
the class i want to spy i have a primary constructor with some dependencies.
So obviously it complained it couldn’t initialize.

I tried doing what i would do in java and made 2 constructors, one with
dependencies, one without and created lateinit var fields for the
dependencies, but then i get errors that the properties have not been
initialized.

So still figuring out how to use the spy with mandatory parameters


On Thu, Nov 23, 2017 at 4:32 PM, Oleksiy Pylypenko <notifications@github.com
> wrote:

> It was added this morning. Try 1.5.4
>
> 23 нояб. 2017 г. 4:29 PM пользователь "Yannick Van Godtsenhoven" <
> notifications@github.com> написал:
>
> > I found this in the docs
> >
> > verify {
> > listOf(obj2, obj3) wasNot Called}
> >
> >
> > but it doesn’t seem to compile
> >
> >
> > On Thu, Nov 23, 2017 at 4:26 PM, Yannick Van Godtsenhoven <

oleksiyp:
Looks like a bug with lateinit var. Fields are copied during spy creation,
just by java reflection, so this one seems copied wrongly. Wonder if there
is easy fix. Because beside that Kotlin support `by` construct, lazy
variables e.t.c. need to dig into that

23 нояб. 2017 г. 4:58 PM пользователь "Yannick Van Godtsenhoven" <
notifications@github.com> написал:

> cool, really like the wasNot Called. nice readability of the tests
>
> was trying out the spyk() but with less success. I also use dagger, so in
> the class i want to spy i have a primary constructor with some
> dependencies.
> So obviously it complained it couldn’t initialize.
>
> I tried doing what i would do in java and made 2 constructors, one with
> dependencies, one without and created lateinit var fields for the
> dependencies, but then i get errors that the properties have not been
> initialized.
>
> So still figuring out how to use the spy with mandatory parameters
>
>
> On Thu, Nov 23, 2017 at 4:32 PM, Oleksiy Pylypenko <
> notifications@github.com
> > wrote:
>
> > It was added this morning. Try 1.5.4
> >
> > 23 нояб. 2017 г. 4:29 PM пользователь "Yannick Van Godtsenhoven" <
> > notifications@github.com> написал:
> >
> > > I found this in the docs
> > >
> > > verify {
> > > listOf(obj2, obj3) wasNot Called}
> > >
> > >
> > > but it doesn’t seem to compile
> > >
> > >
> > > On Thu, Nov 23, 2017 at 4:26 PM, Yannick Van Godtsenhoven <

oleksiyp:
Opened a bug for this issue #7

PR #8, opened 2017-11-25
Corrected excatly to exactly.

jameskbride:
Corrected the documentation for `verify` in the DSL.  (I do think excatly sounds better, from a purely cats-on-the-internet standpoint 😄


oleksiyp:
Thanks. Excatly what I need!

PR #23, opened 2018-02-09
Allow mock creation with a KClass

FredDeschenes:
This allows us to create mocks when interacting with non-Kotlin code
(ex: JUnit5 extensions).

I wanted to create a JUnit5 extension (like [this one](https://github.com/junit-team/junit5-samples/blob/master/junit5-mockito-extension/src/main/java/com/example/mockito/MockitoExtension.java) for Mockito) to handle injecting mocks in function parameters in my tests, but couldn't since JUnit only gives us a `Class`, making it impossible to call the `mockk<T>(...)` function.

### DIFF
--- a/common/src/main/kotlin/io/mockk/MockK.kt
+++ b/common/src/main/kotlin/io/mockk/MockK.kt
@@ -12,8 +12,19 @@ inline fun <reified T : Any> mockk(
     relaxed: Boolean = false,
     vararg moreInterfaces: KClass<*>,
     block: T.() -> Unit = {}
+): T = mockk(T::class, name, relaxed, *moreInterfaces, block = block)
+
+/**
+ * Builds a new mock for specified class
+ */
+inline fun <T : Any> mockk(
+        klazz: KClass<T>,
+        name: String? = null,
+        relaxed: Boolean = false,
+        vararg moreInterfaces: KClass<*>,
+        block: T.() -> Unit = {}
 ): T = MockK.useImpl {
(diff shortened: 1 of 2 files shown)


oleksiyp:
I am very scrupulous about changes to DSL. Alike what am worried is that user in IDEA would need to type few keystrokes more to select one more mockk function. Sorry for that. But really this DSL is for user interaction.

What am I thinking is that this Gateway API layer exactly created for programmatic access. Can you try something alike `useImpl { MockKGateway.implementation().mockFactory.mockk(klazz, name, relaxed, moreInterfaces) }`

FredDeschenes:
It doesn't require any more typing though, the reified functions are still there.

The `MockKGateway...mockk` does work though, so I guess I'll use that instead (I though I needed the `useImpl` stuff (which is internal), but the initialization block isn't necessary in my case). Thanks!

oleksiyp:
Beware that `useImpl` part is not needed because you probably initialized it somewhere else before. I can open it i.e. make public.

oleksiyp:
@FredDeschenes version 1.7.3 released with public `useImpl`

oleksiyp:
@FredDeschenes btw how's progress on your JUnit5 extension? I wonder if you can contribute it to MockK. I would be very pleased if something like that would be included in library itself.

FredDeschenes:
@oleksiyp : I have a simple version that (although highly untested) seems to work fine and cover most use cases. I'll try to make a PR some time this week.

PR #28, opened 2018-02-18
inject mockk implementation supporting mocking fields with attributes…

itsryry:
… of non-same types and only through property injection


oleksiyp:
I looked through the code. I'll do more detailed check in the evening. Thanks

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -55,6 +58,17 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
                     arrayOf()
                 )
             }
+
+            if (property.findAnnotation<InjectMockKs>() != null) {
+                val injectMockProperty = property as KProperty1<Any, Any>
```
oleksiyp: Better keep it semantically named alike `propertyToInjectMockKs`. OR as an option in small scoped variables I use simpler name alike just `property`

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -55,6 +58,17 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+                injectMockProperties.add(injectMockProperty)
+            }
+        }
+
+        if (!injectMockProperties.isEmpty()) {
```
oleksiyp: this `nonEmpty` check does not make any sense as first actual `hasNext` of `for` cycle would do totally same check.

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -55,6 +58,17 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+            for (injectMockProperty in injectMockProperties) {
+                matchInjectMockMembers(target, injectMockProperty)
```
oleksiyp: Naming it `injectMockKs` is okay as well, but up to you. 

From my point of view it is important to keep style at a high level of readability, maybe not all current code is following that, but that means we should make it even better than it's now. Each name should be semantically correct ideally. I prefer small names for small scopes and big names for bigger scopes.

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
oleksiyp: Are you sure that it is only KProperty1? KProperty2 is probably a property with extension or something alike. Please check this case, as it can result in class cast

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -66,6 +80,38 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
         }
     }
 
+    private fun matchInjectMockMembers(target: Any, property: KProperty1<Any, Any>) {
```
oleksiyp: Overall I think we should first learn how Mockito implements this feature. And probably do even better ;-)  

Mockito ***creates*** an instance marked with InjectMocks substituting already created mocks to a constructor. 

Resolution rules are an important part of the design for this task. It should be pretty clear what is happening in case of two Mocks of same types, e.t.c.

unfaded: Thanks for reviewing this PR, didn't want to proceed further until you did a quick check.

Definitely agree on being better than Mockito with the `@InjectMocks` implementation.

`@InjectMocks` seems to be problematic for many with Mockito.  The main problems you'll find when googling Mockito's `@InjectMocks` is how unpredictable their behavior is between same types defined in the test vs the injectable mock target.  They try to match by the property name or the string argument given to `@Mock` annotation (i.e. `@Mock("foo")`).  If they fail, they don't report errors out.  We should definitely be clear on the resolution rules so users do not see unpredictable behavior of same types.  We can also be better with the error handling.

Based on the [documentation](http://static.javadoc.io/org.mockito/mockito-core/2.15.0/org/mockito/InjectMocks.html), Mockito can create the `@InjectMock` property either 2 ways.
- no-arg constructor
- constructor with most parameters (used during construction strategy as opposed to setter or property inject strategies)

If construction strategy fails, setter or property strategy is explored by first creating the instance with zero-arg constructor.  Documentation linked above has good information.

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
itsryry: will remove

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
         }
     }
 
+    private fun resolveInstance(target: Any, property: KProperty1<Any, Any>): Any? {
+
+        var obj: Any?
+        val objCls = property.returnType.classifier as? KClass<*>
+
+        if (objCls != null) {
```
oleksiyp: `if (objCls == nul) throw Exception()` better as make it lower amount of idenation levels

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+            obj = objCls.objectInstance
+
+            if (obj == null) {
+                if (objCls.constructors.isEmpty() && objCls.primaryConstructor == null) {
```
oleksiyp: Isn't `objCls.primaryConstructor == null` always true if `objCls.constructors.isEmpty()` is true?

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+                    throw MockKException("No constructors found to instantiate @InjectMockK field ${property.name}")
+                } else {
+
+                    obj = resolveByConstructorInjection(target, objCls)
+
+                    if (obj == null) {
+                        try {
+                            val noArgCtor = objCls.constructors.first {
```
oleksiyp: Better use firstOrNull and check from my point of view. Again lowering indentation level, as it is really high for this function

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+            throw MockKException("Unable to create @InjectMockK field ${property.name} - could not cast it into a KClass type " +
+                    "required for instance creation.")
+        }
+
+        return obj
+    }
+
+    private fun resolveByConstructorInjection(target: Any, injectMockKCls: KClass<*>) : Any? {
+        // use largest constructor like mockito - maybe we can improve this part
+        val ctor = injectMockKCls.constructors.maxBy { it.parameters.size }
```
oleksiyp: you can `?: throw MockKException("no biggest constructor found")` to get rid of nullable type after

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+        } else {
+            throw MockKException("Unable to create @InjectMockK field ${property.name} - could not cast it into a KClass type " +
+                    "required for instance creation.")
+        }
+
+        return obj
+    }
+
+    private fun resolveByConstructorInjection(target: Any, injectMockKCls: KClass<*>) : Any? {
+        // use largest constructor like mockito - maybe we can improve this part
```
oleksiyp: Yeah, we can try to take into account parameter names, because Kotlin has named parameters.

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+        val ctor = injectMockKCls.constructors.maxBy { it.parameters.size }
+        val targetCls = target::class
+        val noArg = ctor?.parameters?.isEmpty() ?: true
+
+        if (!noArg) {
+            try {
+                val ctorParams = ctor?.parameters
+                val matchingTypes: MutableMap<KParameter, Any?> = mutableMapOf()
+                ctorParams?.forEach { ctorParam ->
+                    targetCls.memberProperties.forEach { testProp ->
```
oleksiyp: Again add filter for class KProperty1

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+                        if (isMatchingType(ctorParam.type, testProp.returnType)) {
+                            matchingTypes[ctorParam] = (testProp as KProperty1<Any, Any?>).get(target)
+                        }
+                    }
+                }
+
+                return ctor?.callBy(matchingTypes)
+            } catch (ex: NoSuchMethodException) {
+                // ignore, attempt property strategy
+                ex.printStackTrace()
```
oleksiyp: Don't get why this exception can happen. Should be handled appropriately

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+                }
+            } else {
+                throw MockKException("No matching attributes found in class $target \n " +
+                        "for @InjectMock field ${obj::class.qualifiedName}")
+            }
+        }
+    }
+
+
+    private fun isMatchingType(t1: KType, t2: KType) : Boolean {
```
oleksiyp: `target` and `object` still short and descriptive, rather than t1, t2

oleksiyp:
Things to consider:

1) Mocks have names, and parameters in Kotlin have names. 
It is possible to specify mock name via annotation (or if not specified it takes name of property AFAIK)
So we can try further resolution by name if two instances of the same type found.

2) Parameters can have defaults. We should allow to skip parameters with defaults I believe

3) Should we consider sub-classes, implemented/extended interfaces when resolving? I don't know how it works in Mockito. Would be pretty weird if yes.

4) Should we take primary constructor instead of max argument count? This would be more Kotlin style, but we still should support Java well. Good to have few Java tests for such kind of cases.

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
+                            (property as KMutableProperty1<Any, Any?>).set(target, obj)
+                        } catch (ex: NoSuchElementException) {
+                            throw MockKException("No suitable constructors found to instantiate @InjectMockK field ${property.name}")
+                        }
+                    } else {
+                        (property as KMutableProperty1<Any, Any?>).set(target, obj)
+                    }
+                }
+            } else {
+                throw MockKException("We could not inject mocks inside of @InjectMockK field ${property.name} because " +
```
oleksiyp: Probably it is not what we want to achieve. As it is possible that user has some code that relies on assigned values. Asking to change because of testing is what we want to avoid actually. Just overwrite values if needed. 

I would even try inject to `vals` via some dirty hack. This would add powerfulness.

mockk/src/main/kotlin/io/mockk/impl/annotations/JvmMockInitializer.kt
```diff
@@ -67,9 +78,113 @@ class JvmMockInitializer(val gateway: MockKGateway) : MockKGateway.MockInitializ
...
 
+    private fun resolveInstance(target: Any, property: KProperty1<Any, Any>): Any? {
+
+        var obj: Any?
+        val objCls = property.returnType.classifier as? KClass<*>
+
+        if (objCls != null) {
+            obj = objCls.objectInstance
+
+            if (obj == null) {
```
oleksiyp: Same about indentation levels

oleksiyp:
btw @itsryry seems it is pretty demanded feature. Just've recently one of the guys were asking it in gitter. I think around 4-5 people were asking so far.

oleksiyp:
already implemented

PR #42, opened 2018-03-12
fix JDK7 compatibility

monosoul:
Now mockk is unusable with JDK7 'cause `Integer.toUnsignedString()` is private in it. This PR is an attempt to fix this incompatibility.


oleksiyp:
Thanks!

monosoul:
It seems like there are a lot more compatibility problems than a thought at first. :disappointed: 
For example: ther's a class called `JvmAutoHinter`, which uses `forEach` method of map. JVM's Map has such method since JDK8 and Kotlin collections also has such method. The problem is: for some reason JVM's implementation is being used => lambda implements interface `java.util.function.BiConsumer` which is of course unavailable in JDK7. I guess it's a Kotlin's compiler plugin problem, cause target bytcode for your project should be 1.6.

Issue #43, opened 2018-03-12
Broken JDK7 compatibility


monosoul:
Mockk 1.7.9 (and all other versions I guess) is currently unable to run under JDK7 due to forEach lambda in `JvmAutoHinter` for example.

monosoul:
@oleksiyp if you don't mind I'd be glad to dig deeper into that problem.

oleksiyp:
Very welcome

monosoul:
@oleksiyp Hey, thanks for merging! :smile: Is there any chance of releasing new version anytime soon?

oleksiyp:
Most probable it will happen today in the evening

monosoul:
Nice, thank you!

monosoul:
Everything is working as expected now. Thanks!

PR #44, opened 2018-03-13
fix JDK7 compatibility (#43)

monosoul:
It took only 2 small changes to bring JDK7 compatibility back. At least functionality that I'm currently using is working perfectly now.

### DIFF
--- a/mockk/src/main/kotlin/io/mockk/impl/instantiation/JvmMockFactoryHelper.kt
+++ b/mockk/src/main/kotlin/io/mockk/impl/instantiation/JvmMockFactoryHelper.kt
@@ -50,26 +50,6 @@ internal object JvmMockFactoryHelper {
         }
     }
 
-
-    fun Method.varArgPosition(): Int {
-        val kFunc =
-            try {
-                // workaround for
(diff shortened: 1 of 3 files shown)


mockk/src/main/kotlin/io/mockk/impl/instantiation/JvmMockFactoryHelper.kt
```diff
@@ -67,7 +67,7 @@ internal object JvmMockFactoryHelper {
                 .filter { it.kind != KParameter.Kind.INSTANCE }
                 .indexOfFirst { it.isVararg }
         else
-            parameters.indexOfFirst { it.isVarArgs }
+            if (isVarArgs) parameterTypes.size - 1 else -1
```
oleksiyp: I have doubts that this is semantically same expressions. You should note that Kotlin varargs can go any place in parameters! Can you please check?

monosoul: Indeed that's so for Kotlin, but we're using Java's `Method` here. I've moved this method out of `JvmMockFactoryHelper ` and introduced a simple test for it.

Issue #58, opened 2018-04-08
Support Kotlin/Native in MockK


olonho:
For multiplatform projects it would be really nice if MockK would support Kotlin/Native as well. If some missing functionality on K/N side is needed, please let us know, and we will add it.

olonho:
Runtime instrumentation is indeed likely out of the question. Unless we will do something like https://en.wikipedia.org/wiki/DTrace with lightweight probes. 

Compile-time wise, there are options. Most obvious one is a source transformation, but it isn't really nice. Another one is to implement compiler plugins API, and do something like https://github.com/JetBrains/kotlin-native/blob/master/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/TestProcessor.kt on IR. 

I think plugin compiler API is most sensible option, we can try to implement an extension point which runs an additional pass, if certain plugin is available. Do you find acceptable option to implement simple IR transformer?

oleksiyp:
Yes, IR transformer seems to be the way to go. 

One more possible solution is to dynamically intercept functions after code generation i.e. embedding some assembly code on the fly by using additional information that compiler provides and dispatching only needed functions. But this seems to be hard. (I believe it is similar to DTrace, just another level)

Many details and design decision to be discussed if IR transformer is selected.

I'll sketch just a few:
 * mocking can be done to any class in runtime. should we instrument all functions? what can be done better?
 * need instrumentation that can call original functions and return result instead of it. how to do that?
 * how to achieve minimum performance impact?
 * should compiler build transformed version only for testing? is there such possibility?
 * should be supported other mocking frameworks? sooner or later such interference may appear.

oleksiyp:
In general, I can admit that most of the functions of `mockk` itself are transferred to common modules. Besides implementing the interface for function interception there are no hard tasks to perform for porting it to any platform.

Additionally, I struggle with following problems:
 * JS support is experimental because it works quite badly with default parameters which are not supported by JS Proxy class. It would be nice to have solution for multiplatform mocking at the end.
 * some things users would like to have extra, alike mocking inline functions
 * Ideally I would like to fully support Kotlin for all platforms(Java, JS, Native, Android device testing) and possible features without any Java agents or Proxies, just via the processor, and only for legacy use additional Java agents and Proxies.

olonho:
> mocking can be done to any class in runtime. should we instrument all functions? what can be done better?
[...]

PR #60, opened 2018-04-11
 Kotlin Native MockK changes

oleksiyp:
#58


KotlinCMakeModule/CMakeDetermineKotlinCompiler.cmake
```diff
@@ -0,0 +1,38 @@
+if (NOT CMAKE_Kotlin_COMPILER)
+    #TODO determine Kotlin compiler in not hardcoded way
+    set(Kotlin_BIN_PATH
+            $ENV{HOME}/.konan/kotlin-native-macos-0.3/bin
```
olonho: Probably you'd better use 0.6.2, as 0.3 is heavily obsolete.

mockk-native/src/main/kotlin/io/mockk/impl/recording/NativeSignatureValueGenerator.kt
```diff
@@ -0,0 +1,29 @@
+package io.mockk.impl.recording
+
+import kotlinx.cinterop.*
+import kotlin.reflect.KClass
+
+class NativeSignatureValueGenerator : SignatureValueGenerator {
+    fun rnd(min: Number, max: Number): Double {
+        TODO("random")
```
olonho: You could use `platform.posix.random()` here - shall be available on all relevent platforms.

mockk-native/src/main/kotlin/io/mockk/impl/instantiation/NativeInstantiator.kt
```diff
@@ -0,0 +1,11 @@
+package io.mockk.impl.instantiation
+
+import kotlin.reflect.KClass
+
+class NativeInstantiator(instanceFactoryRegistry: CommonInstanceFactoryRegistry) :
+    AbstractInstantiator(instanceFactoryRegistry) {
+    @Suppress("UNCHECKED_CAST")
+    override fun <T : Any> instantiate(cls: KClass<T>): T {
```
olonho: Do you really need this level of reflection? Is it possible to avoid instantiation by KClass, or it's absolutely critical for MockK design?

mockk-native/src/main/kotlin/io/mockk/impl/instantiation/NativeMockFactory.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T : Any> newProxy(
+        cls: KClass<out T>,
+        moreInterfaces: Array<out KClass<*>>,
+        stub: Stub,
+        useDefaultConstructor: Boolean,
+        instantiate: Boolean
+    ): T {
+        TODO("proxy")
```
olonho: Proxies would be even harder to implement in native world than new by KClass. Could it be avoided?

KotlinCMakeModule/CMakeDetermineKotlinCompiler.cmake
oleksiyp: Yes I use 0.6.2. Dont know why it is here. And this two days I was writing special Lowering processor for call interception. Will submit PR this evening hopefully. Just for a review

mockk-native/src/main/kotlin/io/mockk/impl/instantiation/NativeInstantiator.kt
oleksiyp: To answer this I need to check usages. One usage would be just object generation for purpose of call signing. This can be replaced by arbitary non-object pointers. Because such object should have only reference equals and identity hashcode. Second usage is returning dummy object for relaxed mocks, so that no NPE is thrown. Such objects can be just zeroed. And more usages I need to check. Of course we can think of some workarounds and framework limitations, but lets focus on proxy/interception first

mockk-native/src/main/kotlin/io/mockk/impl/instantiation/NativeMockFactory.kt
oleksiyp: Proxy = empty object + interception. I am experimenting with IR transformation right now for interception. The problem I see is that libraries compiled earlier wont be transformed, so need special distribution of transformed libraries

oleksiyp:
@olonho I am struggling in debugging it. Despite the fact that I tried just follow IR generated by compiler it is throwing error. I spent most time on checking IR and fixed few things. Now it is failing in some devirtualized call site. Seems I need to get even deeper to gdb and bitcode.

Dont know what can be done from your side besides review and help to debug it.

This changes are doing interceptor part. To merge as well instantiation part should be written

oleksiyp:
Closing for now this one

PR #97, opened 2018-06-25
Added neq and nrefEq

Kerooker:
Having the code `eq(x, true)` and `refEq(x, true)` might sometimes be too verbose and boilerplate.

I propose the matcher neq and nrefEq as a shortcut to both.
I believe the ideal would be to type `!eq(x)` and `!refEq(x)` but I couldn't find an alternative overloading operators, maybe in a further release.


oleksiyp:
I will merge it later, before release. This is because changes to README will appear right away.

PR #161, opened 2018-10-28
Adds Language specific syntax highlight to code blocks in documentation

sandjelkovic:
Adds Language specific syntax highlight to code blocks where language was missing in `README.md`.
This way the code snippets will be consistently highlighted instead of some being highlighted and other not.


oleksiyp:
Thanks

Issue #166, opened 2018-10-31
Stack trace when trying to mock abstract classes on JDK 11


checketts:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [X] I am running the latest version (1.8.10.kotlin13)
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed

### Expected Behavior

In my code I'm mocking `java.time.Clock` which is an abstract class. I expect that mocking to happen without problems.

### Current Behavior

My test passes, however a large stacktrace is generated: (see below)

This is due to mockK hardcoding the ASM Opcode to ASM5 at https://github.com/mockk/mockk/blob/master/agent/jvm/src/main/kotlin/io/mockk/proxy/jvm/transformation/FixParameterNamesVisitor.kt#L45

#### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. When running on JDK 11
2. Mock out `java.time.Clock` via `every { clock.instant() } returns Instant.ofEpochMilli(12345678)`
3. The test passes (and mock seems to work), but stack trace is created

#### Context

* MockK version: 1.8.10.kotlin13
* OS: OSX
* Kotlin version: 1.3
* JDK version: JDK 11
* JUnit version: 5.3.1
* Type of test: unit test (Spring Extension)

#### Stack trace

```
java.lang.UnsupportedOperationException: This feature requires ASM7
	at net.bytebuddy.jar.asm.ClassVisitor.visitNestMember(ClassVisitor.java:236)
	at net.bytebuddy.jar.asm.ClassVisitor.visitNestMember(ClassVisitor.java:239)
	at net.bytebuddy.dynamic.scaffold.TypeWriter$Default$ForInlining$WithFullProcessing$RedefinitionClassVisitor.onVisitNestMember(TypeWriter.java:4525)
	at net.bytebuddy.utility.visitor.MetadataAwareClassVisitor.visitNestMember(MetadataAwareClassVisitor.java:190)
	at net.bytebuddy.jar.asm.ClassReader.accept(ClassReader.java:651)
	at net.bytebuddy.jar.asm.ClassReader.accept(ClassReader.java:391)
	at net.bytebuddy.dynamic.scaffold.TypeWriter$Default$ForInlining.create(TypeWriter.java:3393)
	at net.bytebuddy.dynamic.scaffold.TypeWriter$Default.make(TypeWriter.java:1930)
[...]

PR #167, opened 2018-10-31
Use OpenReader.ASM_API so we match bytebuddies ASM level

checketts:
Addresses #166 

This mimics Mockito's updates to this class: https://github.com/mockito/mockito/blame/release/2.x/src/main/java/org/mockito/internal/creation/bytebuddy/InlineBytecodeGenerator.java#L251


oleksiyp:
Thanks. Let's wait for tests to pass. I think it is better to test older JDKs before releasing as well

PR #183, opened 2018-11-23
Improve Issue99Test

erikhuizinga:
The tested behaviour was incorrect in the sense that it tested if wasn't `123L` by chance. However, what this needs to test is that after unmocking the time doesn't equal the mocked time any longer.


oleksiyp:
Thanks. You are right.

PR #186, opened 2018-11-24
Add extra article and better example of capturing a list

yannickdeturck:
I've added an extra article that I've written a while ago on Mockito and Kotlin.
I've also updated the example on capturing a list of arguments as there was an error in the same and because nothing was done to show that the list contained the two captured arguments.

Error:
``` kotlin
every {
   obj.recordTelemetry(
     speed = capture(list),
    direction = Direction.SOUTH
  )
} answers {
  println(list.captured()) // probably copy paste error from the previous sample?
   Outcome.RECORDED
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Table of contents:
  - [kotlin-fullstack-sample](https://github.com/Kotlin/kotlin-fullstack-sample/pull/28/files#diff-eade18fbfd0abfb6338dbfa647b3215dR17) project covered with tests
  - [DZone article](https://dzone.com/articles/new-mocking-tool-for-kotlin-an-alternative-to-java)
  - [Habrahabr article](https://habrahabr.ru/post/341202/) (RU)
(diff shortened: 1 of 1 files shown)


oleksiyp:
The example is valid as there is a special shortcut function in that `answers` scope:
```
    inline fun <T> MutableList<T>.captured() = last()
```

yannickdeturck:
> The example is valid as there is a special shortcut function in that `answers` scope:
> 
> ```
>     inline fun <T> MutableList<T>.captured() = last()
> ```

Ahh, wasn't aware of that. I tried to use `list.captured` outside the `answers` scope.
I'll revert that change and we'll just keep the article.

oleksiyp:
@yannickdeturck btw can you mask version `1.7.1` in the article and insert badge https://img.shields.io/maven-central/v/io.mockk/mockk.svg?label=release somewhere near? (if it is not insertable then just add a small note to suggest to go to mockk.io for recent version).

yannickdeturck:
Sure, like this?

image

oleksiyp:
Yep

yannickdeturck:
Pushed the change. :)

Issue #240, opened 2019-01-29
autohinting may fail unexpectedly when run in mixed mode, suppressing the useful message that a hint might be necessary in the test fixture


ursjoss:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

If autohinting fails, MockK should indicate that manually setting a hint might be a good idea.

### Current Behavior

If the tests run in mixed mode, type erasure for return types may result in ClassCastExceptions with a message being `null`. This is currently not expected (using `!!` in the method `extractClassName`  in the `JvmAutoHinter`) and results in a `KotlinNullPointerException` instead of simply rethrowing the `ClassCastException` that would later result in the hint about hints :-) being logged.

#### Steps to Reproduce

Unfortunately I have no simple reproducer that reliably illustrates the failing scenario. But looking at the code is already quite convincing given the fact that null messages in ClassCastExceptions may occur in the wild.

#### Context

I have a number of Spek tests using MockK for mocking. All are running successfully when run via IntelliJ spek framework plugin. Two out of those tests keep failing when run through maven, but only on the build server, not reliably on my local development machine. This seems to have to do with running in mixed mode (see e.g.  https://stackoverflow.com/questions/40502576/java-lang-classcastexception-with-null-message-and-cause/40508218).

* MockK version: 1.9
* OS: Windows 10
* Kotlin version: 1.3.20
* JDK version: 1.8.0_192-b12
* Spek Version: 2.0.0-rc1
* Type of test: unit test

#### Stack trace

```
...
kotlin.KotlinNullPointerException
	at io.mockk.impl.recording.JvmAutoHinter.extractClassName(JvmAutoHinter.kt:47)
	at io.mockk.impl.recording.JvmAutoHinter.autoHint(JvmAutoHinter.kt:27)
	at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:36)
	at io.mockk.impl.eval.EveryBlockEvaluator.every(EveryBlockEvaluator.kt:25)
	at io.mockk.MockKDsl.internalEvery(API.kt:92)
	at io.mockk.MockKKt.every(MockK.kt:104)
...
```

I pull request will follow immediatly.

oleksiyp:
Oh really? this NPE is tricky

oleksiyp:
[...]

PR #241, opened 2019-01-29
JvmAutoHinter to cope with ClassCastExceptions with a message being null

ursjoss:
* The ClassCastException thrown in case of missing hints for erased
  return types may have a null message when running in mixed mode.
* The code assumed the message can never be null. If it was null
  nevertheless, a KotlinNullPointerException was thrown (instead of the
  expected ClassCastException that should have been rethrown and would
  have been translated into a useful message that a hint might be
  necessary.)

This may solve or provide input for solving #240.

### DIFF
--- a/mockk/jvm/src/main/kotlin/io/mockk/impl/recording/JvmAutoHinter.kt
+++ b/mockk/jvm/src/main/kotlin/io/mockk/impl/recording/JvmAutoHinter.kt
@@ -44,7 +44,7 @@ class JvmAutoHinter : AutoHinter() {
     }
 
     private fun extractClassName(ex: ClassCastException): String? {
-        return exceptionMessage.find(ex.message!!)?.groups?.get(3)?.value
+        return ex.message?.let { exceptionMessage.find(it)?.groups?.get(3)?.value }
     }
 
     companion object {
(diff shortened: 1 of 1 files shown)


ursjoss:
Running MockK with this patch on our build-server (with the failing Spek tests still without the necessary hints), the build will still fail, but instead of throwing a `KotlinNullPointerException`, a more useful stacktrace is logged, in my oppinion what should be expected:

```
io.mockk.MockKException: 
Class cast exception. Probably type information was erased.
In this case use `hint` before call to specify exact return type of a method. 
	at io.mockk.impl.InternalPlatform.prettifyRecordingException(InternalPlatform.kt:75)
	at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:62)
	at io.mockk.impl.eval.EveryBlockEvaluator.every(EveryBlockEvaluator.kt:25)
	at io.mockk.MockKDsl.internalEvery(API.kt:92)
	at io.mockk.MockKKt.every(MockK.kt:104)
```

oleksiyp:
Oh, great PR. Need to check it in details. Most probably during werkend

ursjoss:
Thanks @oleksiyp . Looking forward to seeing the final resolution to this issue.

And thanks for MockK in general! I like working with it.

ursjoss:
After a good night's sleep I figured we can directly rethrow the message-less ClassCastException, thus skipping the second attempt of getting the class name. It would not be able to do anything more.

There was no noticable improvement of the build time of the project though, but only comparing one build with or without this latest commit each.

ursjoss:
@oleksiyp Do you need anything from me on this one?

oleksiyp:
Not yet. Hopefully will switch to mockk things soon

ursjoss:
@oleksiyp thanks for merging.

PR #303, opened 2019-05-21
Updated readme with version info tags

nisrulz:
The version tag earlier was not quite accurate as it shows the latest version for the dependency being the one that is compatible with Koltin 1.2 and nothing about the one that is compatible with Kotlin1.3+

Just made it clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,6 @@ Table of contents:
 All you need to get started is just to add a dependency to `MockK` library.
 
 #### Gradle/maven dependency
-
 <table>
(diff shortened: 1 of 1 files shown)


oleksiyp:
Tbh Im too lazy to update it manually at each release. I would do some serverless function to extract yhis versions.

nisrulz:
Ideally, these two versions should have been two different artifacts. I am not quite sure how do you even maintain and release two of these from the same repo?

oleksiyp:
It is not very complex. I gave it another thought and it is better to have your changes, rather than. Can save somebody some time

Issue #193, opened 2018-12-03
Feature: switch mainline version to use Kotlin 1.3


PaulWoitaschek:
Could you use standard semver versioning?

I use the [Gradle Versions Plugin](https://github.com/ben-manes/gradle-versions-plugin) and it thinks that .kotlin13 is an outdated version.

I think more consistent would be to release a different artifact so it would be
`io.mockk:mockk-kotlin13:1.8.13` instead of `io.mockk:mockk:1.8.13.kotlin13`

oleksiyp:
Hi, releasing other artifact is not feasable. Due to few reasons. Propose something other and will see :smiley:

PaulWoitaschek:
Remove the kotlin13 suffix and prefix the version with the outdated kotlin version as preKotlin13?

When setting up mockk initially I got confused anyways because I thought the default artifact wouldn't have an outdated kotlin version.

oleksiyp:
You are right it's time to switch kotlin13. So that old one is `1.8.14.kotlin12` and new one `1.8.14` that's not so bad idea I think. Any disadvanteges of that except again some confusion(which Ill try to mitigate by adding a note to the main page at the top)

PaulWoitaschek:
I don't think there are other disadvantages. I assume mockk users are progressive devs anyways and keep their kotlin version up to date anyways.

oleksiyp:
Version 1.9 is released. It is a version using new Kotlin and coroutines. Version 1.9.kotlin12 is kept for compatibility

Issue #151, opened 2018-09-27
 Kotlin 1.3 support 


MrVilkaman:
Please add  support kotlin 1.3 and not experimental coroutines!

Now I use:
`org.jetbrains.kotlinx:kotlinx-coroutines-android:0.26.1-eap13`
`io.mockk:mockk:1.8.8`
`kotlin_version = '1.3.0-rc-57`

I got error  when try use `coEvery{}`

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

io.mockk.MockKException: Add coroutines support artifact 'org.jetbrains.kotlinx:kotlinx-coroutines-core' to your project 

	at io.mockk.impl.InternalPlatform.prettifyRecordingException(InternalPlatform.kt:93)
	at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:50)
	at io.mockk.impl.eval.EveryBlockEvaluator.every(EveryBlockEvaluator.kt:25)
	at io.mockk.MockKDsl.internalCoEvery(API.kt:99)
	at io.mockk.MockKKt.coEvery(MockK.kt:115)
	at ru.skyeng.kids.common.core.action.login.LoginByEmailImplTest$success login by email and pass$1.invokeSuspend(LoginByEmailImplTest.kt:38)
	at ru.skyeng.kids.common.core.action.login.LoginByEmailImplTest$success login by email and pass$1.invoke(LoginByEmailImplTest.kt)
	at tools.TestTools$runTest$1.invokeSuspend(TestTools.kt:6)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:32)
	at kotlinx.coroutines.DispatchedTask$DefaultImpls.run(Dispatched.kt:146)
	at kotlinx.coroutines.DispatchedContinuation.run(Dispatched.kt:13)
	at kotlinx.coroutines.EventLoopBase.processNextEvent(EventLoop.kt:140)
	at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:70)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:46)
	at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source)
[...]

PR #307, opened 2019-05-29
Kotlin1331

saschpe:
Update to Kotlin 1.3.31 and Coroutines 1.2.1

The Gradle plugin and library dependencies should match, thus remove
separate kotlin_gradle_version field in Gradle files.

Related to issues #151 and #193.


saschpe:
The second commit only touches a few optimizations and style improvements. There's much more possible but I wanted to have a discussion first on what you consider appropriate.

oleksiyp:
Isn't `kotlin13` branch is something you are looking for?

saschpe:
You meant I should submit against it instead? It's otherwise empty, no?

oleksiyp:
What I see is that you convert here from experimental coroutines to regular. Which make big sense as it is already done in `kotlin13` branch.

saschpe:
Ok, I can put those changes on top of `kotlin13`, that branch uses an RC of an obsolete 1.3 version and depends on the EAP maven repo and seems to need a rebase as well ;-)

oleksiyp:
It is more trickier than that :smile: You basically will need to split your changes to things end up in both branches and one for kotlin13.

PR #314, opened 2019-06-12
Change then to than

abangfadli:
Grammatical error. 
"Then" should be used when comparing two things (actual calls vs demanded calls)


oleksiyp:
Thanks

Issue #297, opened 2019-05-09
Bug: MockKAgentException when creating a mock on >= Api level 28 due to missing 64 binaries


jtrollkarl:
### Expected Behavior

Expecting mocked instances to be initialised correctly.

### Current Behavior

Currently, whenever a mock of a class is created a MockKAgentException is thrown (see stacktrace below) on api level 28 or greater. I don't seem to be having any issues on lower API levels.

#### Steps to Reproduce

1. Create a test case containing code to create a mock
2. Run test

#### Context

* MockK version: 1.9.3
* OS: >= API level 28
* Kotlin version: 1.3.11
* JDK version: 8
* JUnit version: 4.12
* Type of test: instrumented test

#### Stack trace

```
Caused by: io.mockk.proxy.MockKAgentException: MockK could not self-attach a jvmti agent to the current VM. This feature is required for inline mocking.
This error occured due to an I/O error during the creation of this agent: java.io.IOException: No such file or directory

Potentially, the current VM does not support the jvmti API correctly
at io.mockk.proxy.android.AndroidMockKAgentFactory.init(AndroidMockKAgentFactory.kt:63)
at io.mockk.impl.JvmMockKGateway.<init>(JvmMockKGateway.kt:46)
at io.mockk.impl.JvmMockKGateway.<clinit>(JvmMockKGateway.kt:172)
... 33 more
Caused by: java.io.IOException: No such file or directory
at java.io.UnixFileSystem.createFileExclusively0(Native Method)
at java.io.UnixFileSystem.createFileExclusively(UnixFileSystem.java:281)
at java.io.File.createTempFile(File.java:2018)
at java.io.File.createTempFile(File.java:2064)
at io.mockk.proxy.android.JvmtiAgent.appendToBootstrapClassLoaderSearch(JvmtiAgent.kt:53)
at io.mockk.proxy.android.AndroidMockKAgentFactory.init(AndroidMockKAgentFactory.kt:48)
... 35 more
```

#### Minimal reproducible code (the gist of this issue)

```kotlin
// -----------------------[ GRADLE DEFINITIONS ] -----------------------
dependencies {
    androidTestImplementation "io.mockk:mockk-android:1.9.3"
}
// -----------------------[ YOUR CODE STARTS HERE ] -----------------------
object MockHelper {
    fun mockIsEnabled(enabled: Boolean = false) {

        val mock: LocationChecker = mockk(relaxed = true) //failure here

        val component = DaggerTestAppComponent
            .builder()
[...]

PR #335, opened 2019-07-19
add x86_64 and arm64-v8a mockkjvmtiagent so

ryansgot:
@oleksiyp This should be enough to get me past this issue as it results in the creation of binaries that target the same architecture that I do: https://github.com/mockk/mockk/issues/297

I'm totally unaware of your policy regarding accepting community contributions and how you like to perform releases, so I'm willing to make updates as necessary.

### DIFF
--- a/agent/android/build.gradle
+++ b/agent/android/build.gradle
@@ -63,6 +63,10 @@ android {
         versionName project['version']
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
         testInstrumentationRunnerArgument "notAnnotation", "io.mockk.test.SkipInstrumentedAndroidTest"
+
+        ndk {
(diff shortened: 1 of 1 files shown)


vferries:
Did you manage to make it work ? I tried to rebuild using your PR and include the generated jar in my classpath but still got the error (java.io.IOException: Initialization of libmockkjvmtiagent.so returned non-zero value of -3)

oleksiyp:
There is no policy yet because there are almost no contributions. 

I am reviewing things from time to time, but not sure about doing the release. You need me to understand the impact to do release right away, otherwise, things will just accumulate for some time and one happy day I will do the release(with all testing, checks, e.t.c.) 

So this targets. Why there is no x86 by default? And was you able to run tests afterward?

PR #338, opened 2019-07-29
Fix pom scm.url

ikesyo:
http://www.github.com is redirected to https://github.com


ikesyo:
Could someone take a look?

Issue #346, opened 2019-08-29
"just runs" DSL addition for private methods


saschpe:
### Expected Behavior
The following should compile:

`every { fragment["initToolbar"]() } just runs`

### Current Behavior
Compile error due to missing overload of the *MockKStubScope.just* extension function:
`Type mismatch: inferred type is MockKStubScope<Any?, Any?> but MockKStubScope<Unit, Unit> was expected`

### Current Workaround
`every { fragment["initToolbar"]() } answers(ConstantAnswer(Unit))` or `every { fragment["initToolbar"]() as Unit} just runs`. The latter is acceptable but a more concise version would be good.

PR #347, opened 2019-08-29
Provide invokeReturnsUnit for private functions returning Unit

saschpe:
Convenience syntax as an alternative to any of the following:

- `every { mock["privateMethod"]() } answers(ConstantAnswer(Unit))`
- `every { mock["privateMethod"]() as Unit} just runs`.

Resolves issue #346.

### DIFF
--- a/README.md
+++ b/README.md
@@ -881,7 +881,8 @@ Additionally more verbose syntax allows to get and set properties, do same dynam
 val mock = spyk(Team(), recordPrivateCalls = true)
 
 every { mock getProperty "speed" } returns 33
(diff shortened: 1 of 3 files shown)


oleksiyp:
Invoke no return may mean that it never returns. Alike System.exit. Please consider renaming it

saschpe:
> Invoke no return may mean that it never returns. Alike System.exit. Please consider renaming it

How about *invokeNoReturnValue*? It's a little longer though...

oleksiyp:
Also not exactly the case as functions in Kotlin always have return type

saschpe:
*invokeReturnsUnit* then?

Issue #323, opened 2019-06-28
Bug: withNullableArg { code } doesn't work


OlSom:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

withNullableArg { code } should work.

### Current Behavior

withNullableArg { code } doesn't work if the argument is null. The code block is executed, but then the verification fails despite 'positive' result.

#### Steps to Reproduce

Run the test below

#### Context

* MockK version: 1.9.3
* OS: Ubuntu 18.04.2 LTS
* Kotlin version: 1.3.40
* JDK version: 1.8
* JUnit version: 4.2
* Type of test: Unit test

#### Stack trace

```
Verification failed: call 1 of 1: MockedClass(#1).test(matcher<String>())). Only one matching call to MockedClass(#1)/test(String) happened, but arguments are not matching:
[0]: argument: null, matcher: matcher<String>(), result: +

Stack trace:
                                                          io.mockk.impl.InternalPlatform.captureStackTrace (InternalPlatform.kt:121)                  
                                                            io.mockk.impl.stub.MockKStub.handleInvocation  (MockKStub.kt:247)                         
                          io.mockk.impl.instantiation.JvmMockFactoryHelper$mockHandler$1.invocation        (JvmMockFactoryHelper.kt:25)               
                                                   io.mockk.proxy.jvm.advice.Interceptor.call              (Interceptor.kt:20)                        
                                                                             MockedClass.test              (TestFile.kt:4)                            
                                                                             TestedClass.test              (TestFile.kt:10)                           
                                                                         TestedClassTest.justTest          (TestFileTest.kt:15)                       
                                                    sun.reflect.NativeMethodAccessorImpl.invoke0           (NativeMethodAccessorImpl.java:-2)N        
                                                    sun.reflect.NativeMethodAccessorImpl.invoke            (NativeMethodAccessorImpl.java:62)         
                                                sun.reflect.DelegatingMethodAccessorImpl.invoke            (DelegatingMethodAccessorImpl.java:43)     
                                                                java.lang.reflect.Method.invoke            (Method.java:498)                          
                                               org.junit.runners.model.FrameworkMethod$1.runReflectiveCall (FrameworkMethod.java:50)                  
                                     org.junit.internal.runners.model.ReflectiveCallable.run               (ReflectiveCallable.java:12)               
[...]

PR #348, opened 2019-08-31
Fix 'withNullableArg' function to make it work properly with null values (#323)

vjgarciag96:
As reported on issue #323, the function `withNullableArg` is not working properly when receiving a null value. 

The problem comes from `TypedMatcher` interface and its `checkType` method. When `InvocationMatcher` executes its `match` method and the matcher is a `TypedMatcher`, the mentioned `checkType` method is executed and makes the invocation matcher to return false for any null argument. 
````
fun checkType(arg: Any?): Boolean {
    if (argumentType.simpleName === null) {
        return true
    }

    return argumentType.isInstance(arg)
}
````

`KClass.isInstance` method is going to always return false if the arg's value is null. Therefore we have a problem with `TypedMatcher` and null values.

The proposed solution consists of creating a new `FunctionMatcher` for nullable arguments (`FunctionWithNullableArgMatcher`). This new matcher has the same behavior as `FunctionMatcher` but it overrides the `checkType` function of `TypedMatcher` to take into account the problem with nullable args. Moreover, `FunctionMatcher` has been slightly modified and now the `matchingFunc` attribute doesn't support nullable types (not needed anymore due to the new matcher).

### DIFF
--- a/dsl/common/src/main/kotlin/io/mockk/API.kt
+++ b/dsl/common/src/main/kotlin/io/mockk/API.kt
@@ -662,15 +662,11 @@ open class MockKMatcherScope(
         return callRecorder.matcher(matcher, T::class)
     }
 
-    inline fun <reified T : Any> match(noinline matcher: (T) -> Boolean): T = matchNullable {
-        when (it) {
-            null -> false
-            else -> matcher(it)
-        }
-    }
+    inline fun <reified T : Any> match(noinline matcher: (T) -> Boolean): T =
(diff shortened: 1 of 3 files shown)


oleksiyp:
Hi, thanks for PR! For this particular one, I need to check details. Is there any urgency for these changes?

vjgarciag96:
There is no urgency for these changes, so no hurries and check them whenever you want

PR #393, opened 2019-11-26
Add clarifying comments for Capturing

kibertoad:


README.md
```diff
@@ -487,8 +487,8 @@ val list = mutableListOf<Double>()
 
 every {
   obj.recordTelemetry(
-    speed = capture(slot),
-    direction = Direction.NORTH
+    speed = capture(slot), // makes mock match call with any value for `speed` and record it in a slot
+    direction = Direction.NORTH // makes mock and capturing only match calls with specific `direction`. Use `any()` to match calls with any `direction`
   )
 } answers {
   println(slot.captured)
```
kibertoad: One additional thing I would like to clarify in this PR is usage of `Outcome.RECORDED` that is not explained anywhere in documentation - but unfortunately I don't understand what it is there for either; would appreciate if someone could explain it to me, so that I could document it.

oleksiyp: It's just part of imaginary example

PR #410, opened 2020-02-07
Add missing documentation for the verify method parameters

monosoul:
This PR adds missing documentation for the following `verify()` method parameters: `timeout`, `verifyBlock`.

The most important is the `timeout`'s doc as it's confusing not to know what time unit is used for it. :smile:


oleksiyp:
@monosoul I try to put information about units always :smile: what a shame it's missing here.

Issue #351, opened 2019-09-10
Bug: mocking Android SDK objects with hidden methods is broken for Android P and newer


fkirc:
**Background**
The painful interaction of Android P mocking and "hidden methods" is a known issue.
"Hidden methods" are public methods of the Android SDK, but they are made unavailable via special "security mechanism" of Android P and later.
Hidden methods are very common throughout the Android SDK.
For example, have a look at https://github.com/mockito/mockito/issues/1395 to get an idea of several mocking bugs that are related to hidden methods.
Note that I am only talking about instrumented tests that are running on an actual Android P device. These bugs should not occur for regular Java Unit tests.

**Android P bug**
Mockk has an additional bug that is related to the bugs that are already known.
Specifically, **Mockk triggers a `NoSuchMethodException` whenever some Android SDK code tries to call a hidden method on a mocked SDK object**.

**Expected behaviour**
This NoSuchMethodException should not happen.
Instead, the hidden method should return some default value.
Of course, it is already a serious problem that we are not able to stub or spy on "hidden methods".
Nevertheless, returning a default value is still better than running into Exceptions at arbitrary places within the Android SDK.

**Bug details**
The NoSuchMethodException is provoked by the Mockk Proxy implementation, which performs a reflection access that the Android P runtime does not like.
The following sample stack trace illustrates the bug for a mocked "android.nfc.Tag" object.

> Caused by: java.lang.NoSuchMethodException: android.nfc.Tag.hasTech [int]
>         at java.lang.Class.getDeclaredMethod(Class.java:2050)
>         at io.mockk.proxy.android.MethodDescriptor.getMethod(MethodDescriptor.kt:22) **// mockk provokes the NoSuchMethodException within the proxy implementation!**
>         at io.mockk.proxy.android.advice.Advice.getOrigin(Advice.kt:37)
>         at java.lang.reflect.Method.invoke(Native Method)
>         at io.mockk.proxy.android.AndroidMockKDispatcher.getOrigin(AndroidMockKDispatcher.java:117)
>         at android.nfc.Tag.hasTech(Unknown Source:14) **// hasTech is a hidden method that is called by some Android SDK code.**

**Possible fixes**
I think that it should suffice to simply catch the NoSuchMethodException and then return a default value. This requires a change of the Mockk Proxy implementation.
However, this fix might have bad side effects for calls to actually non-existing methods.

**Future work**
Returning default values for hidden methods fixes this bug, but it is still not a good solution.
[...]

PR #442, opened 2020-05-22
Fix issue 351 by setting hiddenApiExemptions

duffin22:
This PR aims to fix https://github.com/mockk/mockk/issues/351 by taking advantage of `setHiddenApiExemptions` as mentioned in [this stack overflow answer](https://stackoverflow.com/a/55970138).

We're currently using this logic for Android tests as part of our `AndroidJUnitRunner`, but I think it makes sense to bring this into `mockk` as other people seem to be facing the same issue.

I've published these changes as a snapshot to maven local and it seems to do the trick - tested on P and R emulators and a physical Q device. 

@oleksiyp, let me know if you think this belongs somewhere different or if there are any cases you think it'd be good to test.


oleksiyp:
Looks reasonable to me

PR #521, opened 2020-11-04
Fix #214

### LINKED ISSUE #214: Bug: mocking constructor called by object gives inconsistent results
### Description

I'm trying to mock a class which appears as a private property in an object, and which is constructed when the object initializes. The situation is illustrated more or less minimally by the following code excerpt:

```
object ExampleObject {

    private val exampleClass = ExampleClass()
    fun getExampleProperty(): Int = exampleClass.exampleProperty
}

class ExampleClass {

    val exampleProperty: Int = 1
}

@RunWith(JUnit4::class)
class MockkExampleTest {

    @Before fun setup() {
        mockkConstructor(ExampleClass::class)
        every {anyConstructed<ExampleClass>().exampleProperty} returns 0
    }

    @Test fun test1() {
        assertEquals(0, ExampleObject.getExampleProperty())
    }

    @Test fun test2() {
        assertEquals(0, ExampleObject.getExampleProperty())
    }
}
```

### Expected Behavior

`test1()` and `test2()` should both pass or both fail when run in sequence, since they are identical and don't have any obvious side effects.

### Current Behavior

When the two tests are run in sequence, the first to run passes and the second fails.

#### Steps to Reproduce

You can find this code in the form of an Android Studio project at `https://github.com/sglasman/MockkExample`. Running the tests together in Android Studio will reproduce the problem.

ktiniatros:
1. Use the same constructor mock for a class across all tests
2. Clear a mock constructor when unmocking it

Every time mockConstructor is called, the previous mock was canceled and removed from MockKCancellationRegistry.
This cannot work with kotlin object which is a singleton and holds the same references along the app lifecycle.
I try to solve this bug by keeping the same constructor mock (I don't see why not) and I also am clearing answers etc of constructor mock when unmocking it, as it had side effects and I don't see any reason why previous answers should still work since the user asks to be unmocked.

I run all the tests in the same package and they pass. Let me know if I need to test anything else too, to make sure we won't have regressions.

### DIFF
--- a/dsl/common/src/main/kotlin/io/mockk/API.kt
+++ b/dsl/common/src/main/kotlin/io/mockk/API.kt
@@ -508,7 +508,7 @@ object MockKDsl {
 
             MockKCancellationRegistry
                 .subRegistry(MockKCancellationRegistry.Type.CONSTRUCTOR)
-                .cancelPut(it, cancellation)
+                .putIfNotThere(it, cancellation)
         }
     }
 
@@ -517,6 +517,16 @@ object MockKDsl {
      */
     inline fun internalUnmockkConstructor(vararg classes: KClass<*>) {
         classes.forEach {
+            MockKGateway.implementation().constructorMockFactory.clear(
+                    it,
+                    MockKGateway.ClearOptions(
+                            answers = true,
+                            recordedCalls = true,
+                            childMocks = true,
+                            verificationMarks = true,
+                            exclusionRules = true
+                    )
+            )
             MockKCancellationRegistry
                 .subRegistry(MockKCancellationRegistry.Type.CONSTRUCTOR)
                 .cancel(it)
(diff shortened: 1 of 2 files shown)


dsl/common/src/main/kotlin/io/mockk/API.kt
```diff
@@ -2254,6 +2264,13 @@ object MockKCancellationRegistry {
             map[key] = newCancellation
         }
 
+        fun putIfNotThere(key: Any, newCancellation: MockKCancellation) {
```
Raibaz, line 2267: I think you can achieve the same behavior by using the standard library's [putIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#putIfAbsent-K-V-)

Raibaz (changes requested):

dsl/common/src/main/kotlin/io/mockk/API.kt
ktiniatros: This map is a kotlin map, which does not offer `putIfAbsent` like Java 8 map. Is there something similar in kotlin's map?

Raibaz: No, you're right.

Raibaz (approved):

PR #528, opened 2020-11-12
Fix: 343 Added overload for AndThenAnswer.

Jeibniz:
* Added an overload for AndThenAnswer that has the functionality of AndThen.
* Deprecated  AndThen.


Raibaz (approved):

Raibaz:
I've been thinking again about this and I was wondering: shouldn't it be called `andThenAnswers` rather than `andThenAnswer`, given that it's related to `answers`?

Jeibniz:
Thats a good point! Let's rename it to that

PR #551, opened 2020-12-23
Update README.md

chao2zhang:
Correct the language group of the guides


Raibaz:
Thanks a lot for sending this!

oleksiyp:
Wow, I knew that a lot of users of Kotlin & MockK are from Japan, but only now realized this mistake. Sorry for not checking this in translator...

Issue #519, opened 2020-11-01
API 30 crash


TroyEvans1010:
# Description
When using spyk() in a test using API 30, it crashes.
# Prerequisites
- [ ] implementation "io.mockk:mockk-android:1.10.2"
- [ ] emulator: Pixel 3 API 30   *this is important, because the crash did not reproduce on an emulator with API 29.

# Steps
> Start a new Android Studio project (Android Studio Version: 4.0.1)

> Make a MyUtil class:

```kotlin
    class MyUtil {
        fun sayHi(): String {
            return "Hi"
        }
    }
```

> Make a test in androidTest/

```kotlin
    @RunWith(AndroidJUnit4::class)
    class ExampleInstrumentedTest {
        @Test
        fun testMockMyUtil() {
            val mockMyUtil = spyk(MyUtil())
            every { mockMyUtil.sayHi() } returns "MOCK HI"
        }
    }
```

> Run the test

# Expected: Test succeeds
# Actual: Test fails with this stacktrace:

java.lang.ExceptionInInitializerError
at com.example.pg_mockk.ExampleInstrumentedTest.mockMyUtil(ExampleInstrumentedTest.kt:31)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
[...]

PR #562, opened 2021-01-22
Fix issue 519 by not applying setHiddenApiExemptions workaround for API 30+

duffin22:
Proposed fix for https://github.com/mockk/mockk/issues/519.

The workaround introduced in https://github.com/mockk/mockk/pull/442 for allowing [hidden APIs](https://developer.android.com/distribute/best-practices/develop/restrictions-non-sdk-interfaces) to be accessed via reflection no longer works from API 30. This PR ensures that the workaround is not applied in this case

This means that https://github.com/mockk/mockk/issues/351 will occur for API 30 and above, but this would be a smaller issue.

In the long term, to fix https://github.com/mockk/mockk/issues/351 for later APIs, it might be worth trying to integrate https://github.com/ChickenHook/RestrictionBypass which offers another solution.


duffin22:
@oleksiyp @Raibaz does this seem like a reasonable fix?

Raibaz:
LGTM, thanks for taking the time to look into this!

oleksiyp:
@dennisfen thanks for your contribution!

PR #566, opened 2021-02-05
Fix #538: Change exception message when using not mocked class in every block

### LINKED ISSUE #538: Throw an exception when a non-stubbed class is used in an every {} block
It is possible to put inside an `every` block calls to an object that hasn't been stubbed:

```
val foo = SomeClass() // should be mockk<SomeClass>()

every {
    foo.bar()
} returns "baz"
```

It won't work, because `foo` is not a mockk, but it doesn't return a meaningful error message explaining why and how to fix it.

Ideally, an exception should be thrown with a meaningful error message.

mohasaid:
Proposed fix for #538.

# Background

Exception message that is being thrown when a non mocked class is in every block it's not clear.

### Current message

`Missing calls inside every { ... } block.`

### Proposed message

`Cannot use every { ... } block with a non mocked class`

### DIFF
--- a/mockk/common/src/main/kotlin/io/mockk/impl/recording/states/StubbingState.kt
+++ b/mockk/common/src/main/kotlin/io/mockk/impl/recording/states/StubbingState.kt
@@ -11,7 +11,7 @@ class StubbingState(recorder: CommonCallRecorder) : RecordingState(recorder) {
 
     private fun checkMissingCalls() {
         if (recorder.calls.isEmpty()) {
-            throw MockKException("Missing calls inside every { ... } block.")
+            throw MockKException("Cannot use every { ... } block with not mocked class")
         }
     }
 }
--- a/mockk/common/src/main/kotlin/io/mockk/impl/recording/states/StubbingState.kt
+++ b/mockk/common/src/main/kotlin/io/mockk/impl/recording/states/StubbingState.kt
@@ -11,7 +11,7 @@ class StubbingState(recorder: CommonCallRecorder) : RecordingState(recorder) {
 
     private fun checkMissingCalls() {
         if (recorder.calls.isEmpty()) {
-            throw MockKException("Cannot use every { ... } block with not mocked class")
+            throw MockKException("Cannot use every { ... } block with a non mocked class")
         }
     }
 }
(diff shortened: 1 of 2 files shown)


mockk/common/src/main/kotlin/io/mockk/impl/recording/states/StubbingState.kt
```diff
@@ -11,7 +11,7 @@ class StubbingState(recorder: CommonCallRecorder) : RecordingState(recorder) {
 
     private fun checkMissingCalls() {
         if (recorder.calls.isEmpty()) {
-            throw MockKException("Missing calls inside every { ... } block.")
+            throw MockKException("Cannot use every { ... } block with not mocked class")
```
Daniel-Jacob, line 14: I would say 'a non mocked' instead of not mocked.

Daniel-Jacob (changes requested):

Raibaz:
Hi, thanks a lot for this.

I know it's a very rare case, but that MockKException is being thrown also in the case someone mistakenly declares an empty `every` block, and I would like to keep mentioning that in the message.

How about something like "Missing mocked function calls inside every {} block"? 

Do you think it's clearer than the current message in mentioning that `every {}` blocks should contain mocked functions?

mohasaid:
> Hi, thanks a lot for this.
> 
> I know it's a very rare case, but that MockKException is being thrown also in the case someone mistakenly declares an empty `every` block, and I would like to keep mentioning that in the message.
> 
> How about something like "Missing mocked function calls inside every {} block"?
> 
> Do you think it's clearer than the current message in mentioning that `every {}` blocks should contain mocked functions?

I get your point. IMHO the most frequent use case is when someone does not mock the class inside every block, and rarely when they declare empty block. Maybe the message should focus on non mocked class with some indications about empty every block. Something like:
`Cannot use every { ... } block with not mocked class or keeping it empty`?

Raibaz:
Yep you're right, using a non-mocked class seems much more common to me too than having an empty `every` block.

How about "Missing mocked calls inside every { ... } block: make sure the object inside the block is a mock"?

mohasaid:
@Raibaz Sounds good to me, I will update it!

Raibaz:
Thanks a lot!

PR #567, opened 2021-02-12
Bring MockK up to Kotlin 1.4

boswelja:
Updated Kotlin to 1.4.21, and updated Android projects to properly support the change.
The Android Gradle plugin comes with it's own version of kotlin-stdlib as far as I can tell, so in order to keep Kotlin version as consistent as possible I updated that as well. The new version of the Android Gradle plugin also requires at least Android build-tools 30.0.2, so that's been updated as well.
Unfortunately Dokka scrictly supports Kotlin 1.4.20 at the latest, however I don't expect a minor patch to make a difference in doc creation.
I initially wanted to bring this up to Kotlin 1.4.30, but it seems like the Android Gradle plugin isn't quite there yet (it might be in 7.0.0-alpha, but that's too risky for my liking).

### DIFF
--- a/agent/android/build.gradle
+++ b/agent/android/build.gradle
@@ -59,7 +59,7 @@ android {
 
     defaultConfig {
         minSdkVersion 14
-        targetSdkVersion 28
+        targetSdkVersion 30
         versionName project['version']
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
         testInstrumentationRunnerArgument "notAnnotation", "io.mockk.test.SkipInstrumentedAndroidTest"
--- a/agent/common/build.gradle
+++ b/agent/common/build.gradle
@@ -11,10 +11,3 @@ apply from: "${gradles}/upload.gradle"
 dependencies {
     api project(path: ':mockk-agent-api')
 }
-
-compileKotlin {
-    kotlinOptions {
-        apiVersion = '1.3'
-        languageVersion = '1.3'
-    }
-}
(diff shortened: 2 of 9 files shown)


boswelja:
With the bump to Kotlin 1.4, some Kotlin plugins were marked as deprecated, specifically kotlin-platform-js and kotlin-platform-jvm. Since kotlin multiplatform is apparently in alpha, I haven't updated them here, but we may want to look into this in the future.

Raibaz:
Hi, thanks for sending this through; I already tried upgrading to 1.4 in the past, and it turned out that the version of mockk built with Kotlin 1.4 was not backwards compatible to applications being built with Kotlin 1.3.

Did you by any chance test this importing the mockk jar built with Kotlin 1.4 in a Kotlin 1.3 application?

You can do so by setting the `localrepo` gradle property in gradle.properties to the path of your local maven repo and then running `gradle publish`; it will make the SNAPSHOT release you just built available to your local repo and you will be available to include it in another project.

boswelja:
I haven't tried, as none of my projects use Kotlin 1.3. From my understanding, Android Studio is pushing for Kotlin 1.4 already and the Android Gradle plugin is not far from moving to 1.4 too. If you still want me to check for 1.3 support, I can give it a go but I think the best approach is to either release side-by-side 1.4 and 1.3 binaries or start publishing beta/snapshot versions targeting 1.4 until AGP uses 1.4 stable

Raibaz:
This is no longer up to date, as the next version will contain the upgrade to 1.4.

Issue #563, opened 2021-01-25
Mocking of Android SDK objects with hidden methods broken from Android R


duffin22:
This follows on from issues https://github.com/mockk/mockk/issues/351 and https://github.com/mockk/mockk/issues/519. 

With the [restrictions on non-SDK interfaces](https://developer.android.com/distribute/best-practices/develop/restrictions-non-sdk-interfaces) added from Android P, the reflection used when creating mocks of Android SDK objects fails. For Android P and Q, this could be fixed by calling `dalvik.system.VMRuntime#setHiddenApiExemptions` via meta-reflection (added in https://github.com/mockk/mockk/pull/442). This workaround has been closed up by the Android team from Android R as seen in the [source code for VMRuntime](https://android.googlesource.com/platform/libcore/+/master/libart/src/main/java/dalvik/system/VMRuntime.java#70).

### Prerequisites

- [✓] I am running the latest version
- [✓] I checked the documentation and found no answer
- [✓] I checked to make sure that this issue has not already been filed

### Expected Behavior

Android SDK objects can be mocked in instrumented tests.

### Current Behavior

Android instrument tests fail when trying to mock Android SDK objects.

#### Stack trace

```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:606)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1130)
Caused by: java.lang.reflect.InvocationTargetException
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:596)
	... 1 more
Caused by: java.lang.NoSuchMethodException: android.view.View.invalidateInternal [int, int, int, int, boolean, boolean]
	at java.lang.Class.getMethod(Class.java:2072)
	at java.lang.Class.getDeclaredMethod(Class.java:2050)
	at io.mockk.proxy.android.MethodDescriptor.getMethod(MethodDescriptor.kt:22)
```

#### Minimal reproducible code (the gist of this issue)

The issue only reproduces if you are using real Android SDK objects along with mocked ones, for example:

```kotlin
    @get:Rule
    var rule = ActivityScenarioRule(MainActivity::class.java)

    @Test
    fun testFrameLayout() {
        val mockFrameLayout = mockk<FrameLayout>() // this crashes with the above stack trace
    }
[...]

PR #568, opened 2021-02-18
Fix issue 563 by allowing hidden api reflection from MethodDescriptor

duffin22:
Following on from my comments in https://github.com/mockk/mockk/issues/563, I have applied the workaround from dexmaker for `MethodDescriptor` for Android R. This fixes the issue of mocking Android SDK objects in classes that also use real ones.

For this to work, I had to make a couple of additional changes:

1. To make the android tests run locally, I had to remove the back tick method names from tests. I'm not sure why the unit tests in `mockk-common` are being compiled as part of the androidTests in `mockk-android`, but maybe this is just a local issue. I'm happy to drop this change if the tests work elsewhere.
2. The `isOverridden` extension function in `Advice` began to throw with a null `superclass` after I applied the workaround. I worked around this by making the receiver nullable, but I'm not sure if this has any unexpected consequences.

### DIFF
--- a/agent/android/src/main/kotlin/io/mockk/proxy/android/AndroidMockKAgentFactory.kt
+++ b/agent/android/src/main/kotlin/io/mockk/proxy/android/AndroidMockKAgentFactory.kt
@@ -2,6 +2,7 @@
 
 package io.mockk.proxy.android
 
+import android.annotation.SuppressLint
 import android.os.Build
 import io.mockk.proxy.*
 import io.mockk.proxy.android.advice.Advice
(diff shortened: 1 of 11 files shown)


duffin22:
@oleksiyp @Raibaz does this look reasonable?

Raibaz:
Looks ok to me, thanks a lot for sending this through!

Raibaz:
I just sent out #583 that I'd like to merge before releasing a new version.

I'm aiming to release next week, in case there are issues with #583 I'll try making a release with just this change and the upgrade to Kotlin 1.4.

PR #569, opened 2021-02-18
ci: build project using Github actions

milgner:
I had some issues getting tests to run on my local machine so I set up a Github action.
Maybe you're interested in using this in upstream - might need some additional tweaking as this is basically just the default configuration.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -1,0 +1,23 @@
+name: Java CI with Gradle
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


Raibaz:
This is really interesting, I've been discussing with @dottorblaster about the idea of setting up a build matrix for different JDK and Kotlin versions and this is a great first step in that direction.

Thanks a lot!

milgner:
Great to hear ! In case you're interested I'll gladly put together a more sophisticated setup. Apart from adding a build matrix, some dependency caching would probably be in order, too.

Raibaz:
Sure, if you'd like to spend some time in improving this it would be extremely helpful.

Thanks!

PR #570, opened 2021-02-18
Unmock all after each test with JUnit5 extension

### LINKED ISSUE #142: Feature: auto-clear Mockks after tests execution(integration with JUnit)
Sometimes, when mockking in static way (for instance, mocking constructors, mocking static and mocking objects), one may forget to execute a `clearMock` or `unmockkAll` method, and the mocked object context may be leaked to other test cases.

Right now, the current workaround is to unmockk manually

```
@After
fun after() {
    unmockkAll()
}
```

However, having to add this to every `@After` is pretty cumbersome, and when one forgets, the error might be pretty sneaky.

----
Solution for this issue
--

Add an integration to Mockk that executes `unmockkAll` automatically after a test class.

Possibilities:

- Execute it automatically after every test (`@After`) with a hook into JUnit

milgner:
Closes #142 

As explained in https://github.com/mockk/mockk/issues/142#issuecomment-781674022 I think that the associated test needs to be adapted because the tests in `MockKExtensionTest` don't seem to get run at all.

### DIFF
--- a/mockk/jvm/src/main/kotlin/io/mockk/junit5/MockKExtension.kt
+++ b/mockk/jvm/src/main/kotlin/io/mockk/junit5/MockKExtension.kt
@@ -6,10 +6,8 @@ import io.mockk.impl.annotations.MockK
 import io.mockk.impl.annotations.RelaxedMockK
 import io.mockk.impl.annotations.SpyK
 import io.mockk.mockkClass
-import org.junit.jupiter.api.extension.ExtensionContext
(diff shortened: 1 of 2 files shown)


Raibaz:
Thanks a lot for this!

See the comment on the issue re: MockKExtensionTest not being executed.

milgner:
Thank you for fixing the reason why tests weren't being run. I adapted the code and the auto-unmock tests are working as expected (I tried it by inverting the assertion just to be sure).

PR #573, opened 2021-02-19
Add CI test matrix

milgner:
So tests will be run on multiple Java versions. I also added a cache.

Building releases from tags and uploading them should be possible, too. 
I'm a bit concerned about the failing regression test for `Issue3test89Test` - are you perchance aware of that?


Raibaz:
This is great, thanks a lot!

My next step is to add different kotlin versions to the matrix as well

Issue #484, opened 2020-09-04
How to run tests in /jvm folder?


IgorGanapolsky:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [ x] I am running the latest version
- [x ] I checked the documentation and found no answer
- [ x] I checked to make sure that this issue has not already been filed

### Expected Behavior

We should be able to run unit tests in /jvm folder in this project.

### Current Behavior

There is no instructions, and it is unclear how to run the unit tests in /jvm folder.

### Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

#### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Load the project in Intellij
2. Expand **/jvm** folder
3. See that there is no way to run these tests

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: latest master branch
* OS: MacOS
* Kotlin version: latest
* JDK version: 1.8
* JUnit version: 5
* Type of test: unit test OR android instrumented test
Unit test

#### Failure Logs

Please include any relevant log snippets or files here.

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
Error:Internal error: (java.lang.ClassNotFoundException) javax.xml.bind.annotation.XmlSchema
java.lang.ClassNotFoundException: javax.xml.bind.annotation.XmlSchema
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:435)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:589)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	at com.android.repository.api.SchemaModule$SchemaModuleVersion.<init>(SchemaModule.java:157)
	at com.android.repository.api.SchemaModule.<init>(SchemaModule.java:76)
	at com.android.sdklib.repository.AndroidSdkHandler.<clinit>(AndroidSdkHandler.java:82)
	at org.jetbrains.jps.android.AndroidJpsUtil.getAndroidTarget(AndroidJpsUtil.java:405)
	at org.jetbrains.jps.android.AndroidJpsUtil.getAndroidPlatform(AndroidJpsUtil.java:481)
	at org.jetbrains.jps.android.builder.AndroidPreDexBuildTarget.computeRootDescriptors(AndroidPreDexBuildTarget.java:111)
[...]

PR #576, opened 2021-02-27
Improve Android SDK resolution and build

TopHatCroat:
This PR improves Android SDK resolution by preferring whatever is written in `local.properties` if anything, and other wise it will try to pick the SDK location from environmental variables as specified here https://developer.android.com/studio/command-line/variables
I've removed `ndk.dir` as that has been depricated as seen here https://developer.android.com/studio/projects/install-ndk
Lastly, it will not try to build and run/test with Android related modules included if Android SDK is not found and will show an appropriate message in that case.

It's worth noting that some people might miss Wiki page with these instructions as this information is usually written in `README.md`, but I'm not sure if this file is used to automatically generate the content of [mock.io](https://mockk.io). 
This addresses #484 

Install instructions in Wiki should also be updated as such:

## Fresh development setup

1. Clone repo by executing `git clone git@github.com:mockk/mockk.git`
2. [Optional] Install [Android SDK](https://developer.android.com/about/versions/11/setup-sdk) and [Android NDK](https://developer.android.com/studio/projects/install-ndk)
3. Build project by executing `./gradlew build` or use IntelliJ IDEA integrated tools to build and test

To run all tests, execute:
```sh
./gradlew test
```
or to run a specific test:
```sh
./gradlew --rerun-tasks :mock-jvm:test --tests "io.mockk.gh*"
```
This will run all tests of project `mock-jvm` that are in package `op.mockk.gh`
For more rules you can reference [Gradle docs](https://docs.gradle.org/current/userguide/java_testing.html#simple_name_pattern)


Raibaz:
Thanks a lot for sending this through!

PR #582, opened 2021-03-11
Added `mockkStatic` support for `val Any.xxx get()`

gmazzo:
We have found another use case for `mockkStatic`. Basically it's syntax sugar for when you try to mock an extension property like this:
```kotlin
val Int.square get()= this * this
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,6 +803,13 @@ mockkStatic(Obj::extensionFunc)
 ```
 Note that this will mock the whole `pkg.FileKt` class, and not just `extensionFunc`. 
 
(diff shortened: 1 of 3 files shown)


Raibaz:
Nice, thanks a lot for sending this through!

Just a favor to ask: can you please add a mention to this in the documentation?

It's just a matter of adding it in README.md, around line 800 in the section about mocking extension functions should be ok.

gmazzo:
Updated!

Raibaz:
Thanks a lot!

PR #594, opened 2021-04-03
Fix #435: Remove app_name from mockk-android

### LINKED ISSUE #435: app_name string resource should be deleted from mockk-android
mockk/mockk/android/src/main/res/values/strings.xml

contains `app_name` which will override `app_name` in test application in multi-module projects so this should be deleted from the library or use another name because `app_name` is most used resource name by all developers as I know

anesabml:


Raibaz:
@anesabml thanks for sending this through. Can you elaborate a bit on the testing you did after this change?

anesabml:
@Raibaz I just ran the `check` Gradle task and checked the output for warnings, should I introduce some tests in the code base?

Raibaz:
Can you also try running the android test at least on an emulated device please?

Sorry about this, but I know there have been issues with android instrumented tests in the past so I want to be really sure we don't break anything for android users.

Raibaz:
Hey, I just added AITs to the github actions workflow, can you please rebase this on top of master so that it runs them? Thanks a lot and sorry

anesabml:
@Raibaz No need to apologize. Sure I will run the tests on an emulator and rebase the branch.

anesabml:
@Raibaz I run the tests on an emulator and all the tests passed.

Issue #524, opened 2020-11-08
Mock top level functions


carloschameleon:
I started to work with top level functions in Kotlin in my first approach to functional programming and found a way to mock functions that is, I think, effective but wasn't that obvious at first sight.

```kotlin
// This is inside my test file
import top.level.functions.fooFun

mockkStatic(fooFun.javaMethod!!.declaringClass.kotlin)
```

So I was thinking it would be nice to have a function that make it easier to mockk that top level function so you don't have to implement a solution yourself

```kotlin
// This could be a function inside io.mockk.MockK.kt
fun mockkFunction(kfun: KFunction<*>) = mockkStatic(kfun.javaMethod!!.declaringClass.kotlin)
```

The good thing about this approach is that is not a String like when you mockk a file (mockkStatic("my.package.MyFileKt") ) but you pass the name of the method so if you refactor or you change the name of the method tests and compiler will be aware.

If we add that function to MockK.kt the use of this inside a test will be something like this:
```kotlin
import top.level.functions.fooFun
import io.kotest.core.spec.style.FunSpec
import io.mockk.every

class DummyTest : FunSpec({
    mockkFunction(::fooFun)
   
   test("get suggested labels") {
        every { fooFun("foo") } returns "bar"

        // Do your test here
    }
}
```

Endhuine:
There is already a 
```kotlin
fun mockkStatic(vararg functions: KFunction<*>)
```
which you can already use to mock a top-level function in kotlin targeting the jvm

Raibaz:
Yep you're right, I think this can be closed.

Issue #312, opened 2019-06-03
Bug: failure to match method with default String parameter


ghost:
I'm trying to mock a feature that accepts a parameter but the parameter has a default value.

```
class SomeClass {
    fun someMethod(id: String = "nothing"): SomeOtherClass {
        ...    
    } 
}

...

every { someClass.someMethod() } answers { SomeOtherClass() }
```

### Expected Behavior

The mock works with ignoring the parameter value.

### Current Behavior

Mockk throws `io.mockk.MockKException: no aswer found for: ...`

### Workaround

`every { someClass.someMethod(any()) } answers { SomeOtherClass() }`

oleksiyp:
Should be pretty obvious case tested by tests

zenberg:
Can you please check this again?

Raibaz:
This should be already covered by tests, just need to check if it can be reproduced.

Endhuine:
Can't reproduce this issue with the last version of mockk (1.10.6)

Raibaz:
Thanks for investigating this, I think it can be closed.

PR #611, opened 2021-04-28
Feature/test issue 312 507

Endhuine:
Adding test coverage to avoid regression as in issue [312](https://github.com/mockk/mockk/issues/312)
Adding test coverage to avoid regression as in issue [507](https://github.com/mockk/mockk/issues/524)


Raibaz:
Thanks a lot for submitting this!

The reason why the android instrumented tests are failing is that the minSdkVersion was configured to a very old Android version that did not allow test names containing spaces.

I just upgraded it and pushed the change to master, so if you rebase your branch on the current master the build should pass and we can merge this.

Thanks again!

Endhuine:
Seems like the spacing in tests function name is not working, I'll push a solution without spaces.

PR #612, opened 2021-04-28
java 16 - update byte buddy and github ci

andrew-kirkham:
running in a personal project with java 16 i got the following error

```
2021-04-28 11:54:00.140 [Test worker] WARN  i.m.p.j.t.InliningClassTransformer - Failed to transform class java/lang/constant/Constable
java.lang.IllegalArgumentException: Unsupported class file major version 60
```

i worked around it by using this version of bytebuddy in testImplementation

in this pr i update byte buddy to the latest version and add java 16 to github ci. i updated the setup-java action to v2 to support java 16 and set the distro to zulu (preserving the v1 default)

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -10,7 +10,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        java-version: [8, 10, 11, 12, 13, 14, 15]
+        java-version: [8, 10, 11, 12, 13, 14, 15, 16]
         kotlin-version: [1.3.72, 1.4.30]
         kotlin-ir-enabled: [true, false]
         exclude:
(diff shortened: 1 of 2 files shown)


Raibaz:
Hi, thanks a lot for sending this through!

The reason why the builds with java 16 are failing is that they were being executed with gradle 6.3, which was the one previously in use and didn't support java 16.

I just pushed to master the upgrade to gradle 7, so if you rebase your branch on top of master the build should pass and we can merge this.

Thanks again!

andrew-kirkham:
@Raibaz thank you! i just rebased master into this branch

3flex:
I'm also trying to update to Java 16 on some projects.

I'm seeing the same error on an Android build. I believe Android Gradle Plugin is incompatible with Java 16 currently, so Android tests night need to be skipped on this JDK for now.

Issue #614, opened 2021-04-30
ArrayIndexOutOfBoundException in verifyOrder when no call was recorded


dirkbolte:
The following code throws an ArrayIndexOutOfBoundsException in verifyOrder when no call was recorded.

From what I see in `io.mockk.impl.verify.LCSMatchingAlgo:52`, `allCalls` is an empty list, thus `nEdits[-1]` is called.

#### Minimal reproducible code

```kotlin
import io.mockk.mockk
import io.mockk.verifyOrder
import org.junit.jupiter.api.Test

class MockkTest {
  @Test
  fun `verifyOrder does not throw ArrayIndexOutOfBounds`() {
    val mock: Something = mockk(relaxed = true, relaxUnitFun = true)

    verifyOrder {
      mock.doSomething()
    }
  }

  open class Something {
    fun doSomething() {}
  }
}
```

#### Context: 

* MockK version: 1.10.5 (sorry, can't upgrade ATM)
* OS: MacOS 11.3
* Kotlin version: 1.4.20
* JDK version: 11
* JUnit version: 5.7.0

#### Logs and Stack trace

```
INFO: Using test instance lifecycle mode 'PER_CLASS' set via the 'junit.jupiter.testinstance.lifecycle.default' configuration parameter.
16:52:28.484 [main] DEBUG io.mockk.impl.instantiation.AbstractMockFactory - Creating mockk for Something name=#1

java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 0

	at io.mockk.impl.verify.LCSMatchingAlgo.lcs(LCSMatchingAlgo.kt:52)
	at io.mockk.impl.verify.VerificationHelpers.reportCalls$default(VerificationHelpers.kt:61)
	at io.mockk.impl.verify.OrderedCallVerifier$verify$1.invoke(OrderedCallVerifier.kt:28)
	at io.mockk.impl.verify.OrderedCallVerifier$verify$1.invoke(OrderedCallVerifier.kt:12)
	at io.mockk.impl.recording.CommonCallRecorder.safeExec(CommonCallRecorder.kt:72)
	at io.mockk.impl.log.SafeToString.exec(SafeToString.kt:10)
	at io.mockk.impl.verify.OrderedCallVerifier.verify(OrderedCallVerifier.kt:26)
	at io.mockk.impl.recording.states.VerifyingState$recordingDone$outcome$1.invoke(VerifyingState.kt:31)
[...]

PR #615, opened 2021-04-30
fix[614]: ArrayIndexOutOfBoundException in verifyOrder when no call was recorded

Endhuine:
Added safe calls in the `LCSMatchingAlgo` to avoid `ArrayIndexOutOfBoundException` as in #614 .
The output in this scenario is now:
```
 java.lang.AssertionError: Verification failed: fewer calls happened than demanded by order verification sequence. 

    Matchers: 
    Something(#732).doSomething())

    Calls:


```

### DIFF
--- a/mockk/common/src/test/kotlin/io/mockk/gh/Issue614Test.kt
+++ b/mockk/common/src/test/kotlin/io/mockk/gh/Issue614Test.kt
@@ -8,7 +8,7 @@ import kotlin.test.assertFailsWith
 class Issue614Test {
 
     @Test
-    fun `verifyOrder should throw AssertionError if no matching is found`() {
+    fun verifyOrderThrowAssertionErrorIfNoCallHasBeenMade() {
         val mock: Something = mockk(relaxed = true, relaxUnitFun = true)
 
         assertFailsWith<AssertionError> { verifyOrder { mock.doSomething() } }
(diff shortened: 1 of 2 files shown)


mockk/common/src/test/kotlin/io/mockk/gh/Issue614Test.kt
```diff
@@ -0,0 +1,22 @@
...
+import kotlin.test.assertFails
+import kotlin.test.assertSame
+
+class Issue614Test {
+
+    @Test
+    fun `verifyOrder should throw AssertionError if no matching is found`() {
+        val mock: Something = mockk(relaxed = true, relaxUnitFun = true)
+
+        val exception = assertFails { verifyOrder { mock.doSomething() } }
```
Raibaz, line 15: I think you can write this as 
```
assertFailsWith<AssertionError> {
    verifyOrder { mock.doSomething() }
}
```

Which looks a bit better to me.

What do you think?

Endhuine: Nice catch, I'll update it

Raibaz:
Tests with spaces in the names are still failing when being executed as anrdoid instrumented tests.

Can you please switch them to names without spaces, for the time being, while we try to figure out a way to have them pass on the android emulator?

Sorry about this

PR #620, opened 2021-05-05
Reorganize tests

drazen04:
I tried with two issues (29 and 94). I hope is what you ask for. In case i could continue to reorganize remaining issues.


Raibaz:
Thanks for this, it's a good first step. Moving onwards, I'd like to start grouping similar tests in the same classes.

drazen04:
Thanks for the tip. I'll try to move on this way.

PR #621, opened 2021-05-10
Reorganize tests

drazen04:
Hi, I continued to work on reorganizing test and i tried to grouping similar ones, like you suggest.


mockk/jvm/src/test/kotlin/io/mockk/it/MockStaticMethodTest.kt
```diff
@@ -2,13 +2,43 @@ package io.mockk.it
 
 import io.mockk.every
 import io.mockk.mockkStatic
+import io.mockk.unmockkStatic
 import io.mockk.verify
+import org.junit.Test
+import java.time.Instant
 import kotlin.test.Test
```
Endhuine, line 9: You are importing org.junit.Test, this should be removed

mockk/jvm/src/test/kotlin/io/mockk/it/MockStaticMethodTest.kt
```diff
@@ -2,13 +2,43 @@ package io.mockk.it
...
 
 val Int.selfOp get() = this * this
 
-class StaticMockJVMTest {
+class MockStaticMethodTest {
+    /**
+     * github issue #92
+     */
+    @Test
+    fun test() {
```
Endhuine, line 20: I'd suggest using a better name on tests, describing what the test is actually doing.

drazen04: Thank you for suggest, I didn't change method name because it was not required for the issue, but i will.

mockk/jvm/src/test/kotlin/io/mockk/it/MockStaticMethodTest.kt
drazen04: Thanks. It was imported for Issue92Test, i didn't realize that was a kotlin class for same reason.

drazen04:
Thanks for tips. If you have more, let me know.

Raibaz:
Thanks a lot for sending this through and for the improvements!

(and thanks @Endhuine for the review :) )

PR #631, opened 2021-05-25
feature[test]: refactored following issue tests: 31, 36, 47, 48, 51, 70

Endhuine:
As per the title, the tests for issues  31, 36, 47, 48, 51, 70 were refactored to the internal `it` package


mockk/common/src/test/kotlin/io/mockk/it/InjectMocksTest.kt
```diff
@@ -17,7 +21,7 @@ class Issue47Test {
     }
 
     abstract class AbstractBar<T : IFoo> {
-//        @Inject
+        //        @Inject
```
Raibaz, line 24: Can you please remove this while we're here?

Raibaz (approved):
LGTM pending a very minor change

mockk/common/src/test/kotlin/io/mockk/it/CoroutinesTest.kt
```diff
@@ -51,8 +50,33 @@ class CoroutinesTest {
         }
     }
 
+    /**
+     * See issue #48
+     */
+    @Test
+    fun mockPrivateCoroutineCall() {
+        val myClassSpy = spyk(MockPrivateSuspendCls())
```
Raibaz, line 58: I have been porting mockk to kotlin 1.5.0, and when upgrading, this test passes only if the spy is declared as 

```
spyk<MockPrivateSuspendCls>(recordPrivateCalls = true)
```

Can you please change this? It is indeed how this test should be, so IMHO it's wrong for it to pass as it is :)

Raibaz (changes requested):

Raibaz:
Also, if you rebase on master the checks should pass now.

Raibaz (approved):

PR #633, opened 2021-05-27
feat: value class support

qoomon:


agent/jvm/src/main/kotlin/io/mockk/proxy/jvm/advice/ValueClassSupport.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    // get backing field
+    val backingField = this::class.valueField()
+
+    // get boxed value
+    @Suppress("UNCHECKED_CAST")
+    return (backingField as KProperty1<T, *>).get(this)
+}
+
+fun <T : Any> T.boxedClass(): KClass<*> {
```
Endhuine, line 25: Could you add a brief kdoc on this since it's a public api?

agent/jvm/src/main/kotlin/io/mockk/proxy/jvm/advice/ValueClassSupport.kt
```diff
@@ -0,0 +1,69 @@
...
+        // ...with a backing field
+        val backingField = declaredMemberProperties
+            .first { it.name == constructorParameter.name }
+            .apply { isAccessible = true }
+
+        backingField
+    } as KProperty1<out T, *>
+}
+
+fun <T : Any> KClass<T>.isValueClass() = try {
```
Endhuine, line 54: Could you add a brief kdoc on this since it's a public api?

client-tests/jvm/build.gradle
```diff
@@ -17,13 +17,6 @@ apply plugin: 'kotlin'
 sourceCompatibility = JavaVersion.VERSION_1_8
 targetCompatibility = JavaVersion.VERSION_1_8
 
-compileTestKotlin {
```
Raibaz, line 20: Why was this removed? The point of the client-tests module is to ensure backwards compatibility with projects running on Kotlin 1.3.*.

I see the point in removing the same block in other modules, but I thinks it should be kept in client-tests.

qoomon: Okay, I thought it was just an empty configuration block.

Raibaz:
Looks like all the builds running on 1.3.72 and 1.4.30 are failing because maxOrNull has been introduced in 1.5.0.

I'd like to maintain backwards compatibility with both of them for a little while now - and i think reverting the `apiVersion`/`languageVersion` changes should fix this.

I know it gives a warning in 1.5 mentioning that apiVersion 1.3 is going to be removed, but I'd rather bear with the warning for now instead of breaking backwards compatibility.

What do you think?

qoomon:
I'll try this. However how to run test with api version 1.3 and the new `value class` only available in api version 1.5?
I could use `inline class` however this is a kinda bad test than. What do you think?

qoomon:
It's working. I've revered `maxOrNull()` to `max()` and suppress the compile errors by using `@Suppress("DEPRECATION_ERROR")`

peterfigure:
Thanks for this PR, I've been trying it out locally and it's working well, really hope this gets merged soon! 🙏

mockk/common/src/test/kotlin/io/mockk/it/ValueClassTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.mockk.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class ValueClassTest {
+
+    private val mock = mockk<DummyService>()
+
+    @Test
+    fun `value class object as return value`() {
```
Raibaz, line 12: Can you please use single word test names, e.g. valueClassObjectAsReturnValue? Turns out android does not really like backtick-named tests with more than one word :(

Raibaz (changes requested):

mockk/common/src/test/kotlin/io/mockk/it/ValueClassTest.kt
Raibaz: Sorry i missed this earlier, btw :(

qoomon: sure, no problem.

agent/jvm/src/main/kotlin/io/mockk/proxy/jvm/advice/ValueClassSupport.kt
Kantis: Should it be part of public API? Couldn't it be marked as `internal`? It doesn't feel like something a user of mockk expects to find. Just my 2 cents :)

qoomon: @Kantis I'd like to make it internal, however it is used within `mockk-agent-jvm` and `mockk-jvm` module. Any ideas how to solve this?

Kantis: Hmm, actually I think it'll be fine as-is.

Since the declaration lives in `mockk-agent-jvm` and it's included in `mockk-jvm` using `implementation project(":mockk-agent-jvm")`, the public declarations from `mockk-agent-jvm` won't be visible to users of `mockk`.

Only options to mark it as `internal` is to copy the code to both modules or merge the modules.

peterfigure:
Any way to move this PR forward? Very hard to release projects based on Kotlin 1.5 without this 😅

Raibaz:
It's currently still failing because the mockk-android module is lacking the ValueClassSupport class, which in turn is making a lot of tests fail.

@qoomon, do you think we could move it to mockk-common, so that it is also visible to the android module?

qoomon:
Sure give me a minute

qoomon:
@Raibaz it is not that easy cause `ValueClassSupport` depends on `kotlin.reflect.jvm.isAccessible` and  `java.declaredMethods` (this is just needed for kotlin <1.5 polyfill `isValue` method)

qoomon:
@Raibaz why is `mockk-agent-jvm` excluded within `mokk-android` `build.gradle`

```gradle
   api(mockKProject, {
        exclude group: 'io.mockk', module: 'mockk-agent-jvm'
    })
```

Raibaz:
It's excluded because android has its own agent.

As a possible workaround for this, how about providing a dummy implementation of ValueClassSupport for mockk-android, that basically behaves as the polyfilled version for kotlin <1.5?

This will not provide support for value classes in Android, but it would still allow us to merge this.

What do you think?

qoomon:
Sounds reasonable. Is there an easy way to test if it works without the need of a full android environment?

qoomon:
I did some adjustments so it should work now, however I'm not sure if the changes within `mockk-agent-android` are working because I don't have the right environment

Raibaz:
Thanks a lot for taking care of this!

Unfortunately, Github requires manual approval for first-time contributors, so I have to manually trigger the builds which are, effectively, the fastest way to make sure everything works on Android.

Sorry about this :(

qoomon:
it looks like everything seems to work

Raibaz:
Thanks a lot again!

deadeyye:
@Raibaz when will you make new mockk release with this PR included?

Raibaz:
I have a couple of PRs pending that I'd like to merge to release them all together, hopefully by the end of this week or in the next week.

peterfigure:
@Raibaz any news on the release including this PR?

antondudakov:
ship it 🚀

hrach:
Guys, if you wanna help you may use prebuilts via jitpack.io. I am doing so and just discovered an issue, see latest PR. These comment won't speed this up.

schernetsky:
With this change,, I'm seeing unboxing enter an infinite loop (inside `every` block) when mocking methods returning nullable value classes.

On Kotlin 1.5.10 with MockK 1.12.0

PR #643, opened 2021-06-16
Custom any value generator

### LINKED ISSUE #642: Utilizing nullability in relaxed mock
### Expected Behavior

We're migrating from Mockito and one of the biggest differences is that the relaxed mock is returning always a value, even if the method has a nullable return type.

We consider it a nice feature not to redefine methods we do not care and wanna let the code handle the default, which is rather a null than empty object (list/map/0/"").

Would you be open to add this behavior to MockK (I would try to make PR)?

Since this would be a great BC break, it should be probably configurable. Ideally the same way as the relax behavior:
- via function argument (~`mockk(relaxWithNulls = true)`)
- via project-wide configuration

### Current Behavior

```kotlin
        class Foo {
            fun getSomething(): Int? {
                return 1
            }
        }

        val mock = mockk<Foo>(relaxed = true) 
        assertEquals(null, mock.getSomething()) // fails, returns 0 currently
```

hrach:
Closes #642

### DIFF
--- a/agent/android/src/main/kotlin/io/mockk/ValueClassSupport.kt
+++ b/agent/android/src/main/kotlin/io/mockk/ValueClassSupport.kt
@@ -30,10 +30,19 @@ fun <T : Any> T.boxedValue(): Any? {
  * @return class of boxed value, if this is value class, else just class of itself
  */
 fun <T : Any> T.boxedClass(): KClass<*> {
-    if (!this::class.isValueClass()) return this::class
+    return this::class.boxedClass()
+}
+
+/**
+ * Get the KClass of boxed value if this is a value class.
+ *
+ * @return class of boxed value, if this is value class, else just class of itself
+ */
+fun KClass<*>.boxedClass(): KClass<*> {
+    if (!this.isValueClass()) return this
 
     // get backing field
-    val backingField = this::class.valueField()
+    val backingField = this.valueField()
 
     // get boxed value
     return backingField.returnType.classifier as KClass<*>
--- a/agent/android/src/main/kotlin/io/mockk/ValueClassSupport.kt
+++ b/agent/android/src/main/kotlin/io/mockk/ValueClassSupport.kt
@@ -69,7 +69,7 @@ private fun <T : Any> KClass<T>.valueField(): KProperty1<out T, *> {
 
 private fun <T : Any> KClass<T>.isValueClass() = try {
     this.isValue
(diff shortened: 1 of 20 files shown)


hrach:
Hi @Raibaz this is probably ready to go.

- first commits properly reads the nullability
- the second introduces mutability for any value generator and opens the api a bit
- the third commit is a fix from my previous PR changing this, sadly it passed the tests but running in my unit-tests (not-android test) it was somehow missing - adding copies to proper places
- the fourth commit is fixing reflection exception, I was unable to find repro case as this was pretty much complicated code (generics, interface with inheritance, reactive chains)

I'd tried to fix the android runner as discussed here: https://github.com/ReactiveCircus/android-emulator-runner/issues/160 but doesn't seem to work.

Raibaz:
Whatever, merging this and in case the android builds still fail I'll address them on master.

Thanks a lot for all the work you put into this!

hrach:
Thank you a lot! I have to say your maintainer's response time/attitude exceeded my expectations a lot!

cbxp:
Cannot find any documentation on how to substitute own null-returning AnyValueGenerator...

mafo5:
> Cannot find any documentation on how to substitute own null-returning AnyValueGenerator...

The headline is misleading! The PR extends the `relaxed=true` behavior to the AnyValueGenerator as well.

samba-gangineni:
This is still returning non null value. If I am reading correctly, this PR is supposed to return null if the method return type is nullable?

```
class MockkTest : BehaviorSpec() {
    init {

        class Foo {
            fun getSomething(): Int? {
                return 1
            }
        }
        Given("test") {
            When("test") {
                Then("test") {
                    val mock = mockk<Foo>(relaxed = true)
                    mock.getSomething() shouldBe null // fails, returns 0 currentl
                }
            }
        }
    }
}

// Expected null but actual was 0
java.lang.AssertionError: Expected null but actual was 0
```

samba-gangineni:
nvm looks like I need to implement my own valueGenerator

```
class NullableValueGenerator(
        voidInstance: Any
    ) : JvmAnyValueGenerator(voidInstance) {
        override fun anyValue(cls: KClass<*>, isNullable: Boolean, orInstantiateVia: () -> Any?): Any? {
            if (isNullable) return null
            return super.anyValue(cls, isNullable, orInstantiateVia)
        }
    }
    
    
```

PR #641, opened 2021-06-15
implement support for relaxed mock on suspending function [closes #640]

hrach:

PR #644, opened 2021-06-16
fix reading return type for suspending function returning generic type

hrach:
The previous #641 implemetnation was buggy, a fix here.

### DIFF
--- a/mockk/jvm/src/main/kotlin/io/mockk/ValueClassSupport.kt
+++ b/mockk/jvm/src/main/kotlin/io/mockk/ValueClassSupport.kt
@@ -30,10 +30,19 @@ fun <T : Any> T.boxedValue(): Any? {
  * @return class of boxed value, if this is value class, else just class of itself
  */
 fun <T : Any> T.boxedClass(): KClass<*> {
-    if (!this::class.isValueClass()) return this::class
+    return this::class.boxedClass()
+}
+
+/**
+ * Get the KClass of boxed value if this is a value class.
+ *
+ * @return class of boxed value, if this is value class, else just class of itself
+ */
+fun KClass<*>.boxedClass(): KClass<*> {
+    if (!this.isValueClass()) return this
 
     // get backing field
-    val backingField = this::class.valueField()
+    val backingField = this.valueField()
 
     // get boxed value
     return backingField.returnType.classifier as KClass<*>
--- a/mockk/jvm/src/main/kotlin/io/mockk/impl/instantiation/JvmMockFactoryHelper.kt
+++ b/mockk/jvm/src/main/kotlin/io/mockk/impl/instantiation/JvmMockFactoryHelper.kt
@@ -4,17 +4,15 @@ import io.mockk.*
 import io.mockk.impl.InternalPlatform
 import io.mockk.impl.stub.Stub
 import io.mockk.proxy.MockKInvocationHandler
-import java.lang.Class
 import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Method
 import java.lang.reflect.Modifier
-import java.lang.reflect.ParameterizedType
-import java.lang.reflect.WildcardType
 import java.util.concurrent.Callable
 import kotlin.coroutines.Continuation
 import kotlin.reflect.KClass
 import kotlin.reflect.KMutableProperty
 import kotlin.reflect.KParameter
+import kotlin.reflect.KType
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.jvm.javaField
 import kotlin.reflect.jvm.javaMethod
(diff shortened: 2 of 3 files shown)


hrach:
Finally I was able to obtains the stack trace:
```
java.lang.ClassNotFoundException: Invalid name: java.lang.Boolean[]
at java.lang.Class.classForName(Native Method)
at java.lang.Class.forName(Class.java:454)
at java.lang.Class.forName(Class.java:379)
at io.mockk.impl.recording.JvmAutoHinter.autoHint(JvmAutoHinter.kt:35)
at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:40)
at io.mockk.impl.eval.EveryBlockEvaluator.every(EveryBlockEvaluator.kt:30)
at io.mockk.MockKDsl.internalEvery(API.kt:93)
at io.mockk.MockKKt.every(MockK.kt:98)
at io.mockk.it.ArraysTest.booleanObjectArray(ArraysTest.kt:94)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
at org.junit.runners.Suite.runChild(Suite.java:128)
at org.junit.runners.Suite.runChild(Suite.java:27)
at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
at android.support.test.internal.runner.TestExecutor.execute(TestExecutor.java:56)
at android.support.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:384)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2205)
```

hrach:
@Raibaz please don't you have idea what may be the cause here? Or any idea how to debug this? Thx

hrach:
The issue is that from some reason Android isn't able to handle `kotlin.BooleanArray` and only works when the return type resolution returns `kotlin.Array`. I'm totally lost what this affects and how to properly fix it. Any idea please @Raibaz?

I've added a nasty fix 😢. Not sure how else to fix it. We may introduce some expect/actual but not sure it worth.  Mainly because the fix should be probably elsewhere.

Raibaz:
I've seen the Android failures, and I honestly have no idea why it's failing :(

I'm not familiar enough with the Android ecosystem to have a clue about it.

BTW, looks like now there are failures in the JVM module as well.

Thanks for looking through this though!

hrach:
This JVM failure was due to need to unbox value class after merging their support. So 🤞 fixed now.

hrach:
> This JVM failure was due to need to unbox value class after merging their support.

And that's needed because now we read the return type by kotlin reflection and not throguh java reflection which already had the boxed type.

Raibaz:
Great, thanks a lot for fixing everything!

PR #646, opened 2021-06-21
Add optional constructor arguments support for @InjectMockKs

vladkanash:
Currently, when using `@InjectMockKs` annotation with constructor injection, it is required to provide mocked values for all constructor arguments, even if they are optional and have default values (See example below). 
With this PR `@InjectMockks` will still look for mockks for all parameters, including optional, but if nothing is found, default value will be used.

```
class Example (param1: Param1, param2: Param2 = Param2()) {...}

...

@Mockk
lateinit var param1: Param1

@InjectMockks //Will not work since no mockk for param2 is provided
lateinit var example: Example
```


Raibaz:
Nice, thanks a lot for sending this through!

The build failure looks temporary, I'll merge this as soon as it goes green.

PR #648, opened 2021-06-27
rename some `obj` to `car` in readme sample code

tkhs0604:
When I was reading mockk readme document, I found some `obj` are used, which seems incorrect name. So I renamed them to `car`.


Raibaz:
Wow, you're right, thanks a lot for spotting this and sending this through!

PR #663, opened 2021-07-09
remove deprecated jcenter repository

agebhar1:
> JFrog announced the decommission of JCenter. This has the potential to impact
> many Gradle builds. Gradle emits a deprecation warning when jcenter() is used
> in the repository configuration. Consider using mavenCentral(), google() or a
> private maven repository instead.

https://docs.gradle.org/7.0/userguide/upgrading_version_6.html#jcenter_deprecation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,6 @@ buildscript {
     repositories {
         mavenCentral()
         google()
-        jcenter()
     }
     dependencies {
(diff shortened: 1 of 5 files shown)


Raibaz:
This is great, I remember reading about the jcenter deprecation but I never took the time to see if MockK was affected. Thanks for looking into this!

agebhar1:
You're welcome!

PR #664, opened 2021-07-09
use latest Kotlin versions 1.4.31/1.5.10 for tests

agebhar1:


Raibaz:
Nice, thanks for sending this through!

agebhar1:
You're welcome!

PR #666, opened 2021-07-10
run test(s) w/ current build

agebhar1:
I guess the tests ran with 1.8.8-SNAPSHOT :thinking:


Raibaz:
The `matrix` folder is to manually generate a test matrix with different jdk versions.

I think it can safely be removed now that we have github actions to do the same thing.

agebhar1:
I've updated the PR accordingly.

Raibaz:
Thanks a lot!

agebhar1:
My pleasure!

Issue #565, opened 2021-02-05
Mocking map invocation via chained call stopped working


cbxp:
Mockk version: 1.10.5
Kotlin version: 1.4.21
OS: MacOS Big Sur 11.2
JVM: adopt-openjdk-11 version 11.0.10

Java class under test:
```
import java.util.Map;

public interface Bar {
  Map<String, Integer> getAttributes();
}

```

Test:
```
val bar = mockk<Bar>(relaxed = true)
every { bar.attributes["foo"] } returns null
bar.attributes["foo"]
```

Expected (as in version 1.9.3)
`bar.attributes` should return a `Map`

Actual
`NullPointerException` because `bar.attributes` is null

Raibaz:
Does this still happen if you declare your mock as a relaxed mock, i.e. `val bar = mockk<Bar>(relaxed  = true)`?

cbxp:
@Raibaz It still fails. I will update the example

omesquita:
I have a same problem when i use `val bar = mockk<Bar>(relaxed = true) `and `every { bar.list.isEmpty() } returns false`. 
Someone find a solution?

Raibaz:
This looks like a bug in the `AnyValueGenerator`; it should be returning an empty map for `bar.attributes` if the mock is relaxed, yet it seems like it's returning null.

omesquita:
Thank you @Raibaz.  Can you me say how i fix it?

pedrofsn:
Hello guys. Any updates about this issue?

omesquita:
Hi! Anyone have a solution for this issue?

PR #687, opened 2021-08-29
Fix stubbing by chained calls on collections (#565)

AlexKrupa:
## Related links
Issue: https://github.com/mockk/mockk/issues/565
Changes that affected expected behavior: https://github.com/mockk/mockk/pull/387, https://github.com/mockk/mockk/pull/536

## Introduction
Any feedback is appreciated. I'm completely new to the internals of MockK, so please excuse any shortcomings of my solutions.

**I recommend checking commits separately**, because there are alternative solutions. I'll clean those up in case a solution is chosen.

Tests pass for both solutions.

## Explanation

This is an attempt at fixing stubbing by chained calls on collection, such as:
```kotlin
val mock = mockk<SomeClass>()
every { mock.map.get("foo") } returns 0
```

Expected behavior broke when `(Jvm)AnyValueGenerator` started returning concrete instances for collection types. Before those branches existed, the generator would fall back to `orInstantiateVia` lambda, which – when passed from `RecordingState.call` – created a temporary mock on invocation.

### Solution 1: Return temporary mocks for collection types in `RecordingState.call`
Encapsulating the change here helps avoid concerning other consumers of `AnyValueGenerator`, making the scope of the change much smaller than solution 2.

A possible disadvantage is that these collection (or other) types have to be kept in sync between `RecordingState` and `AnyValueGenerator`. Perhaps we could obtain expose those types from `AnyValueGenerator` instead?

### Solution 2: Add flag to configure `AnyValueGenerator`'s behavior
If interfering in `AnyValueGenerator` isn't a concern, we can pass a flag to indicate whether generated objects should be mocks. This adds more complexity than solution 1.

### DIFF
--- a/mockk/common/src/main/kotlin/io/mockk/impl/recording/states/RecordingState.kt
+++ b/mockk/common/src/main/kotlin/io/mockk/impl/recording/states/RecordingState.kt
@@ -62,16 +62,16 @@ abstract class RecordingState(recorder: CommonCallRecorder) : CallRecordingState
     override fun call(invocation: Invocation): Any? {
         val retType = recorder.childHinter.nextChildType { invocation.method.returnType }
         var isTemporaryMock = false
-
-        val retValue =
-            if (invocation.method.isToString()) {
-                recorder.stubRepo[invocation.self]?.toStr() ?: ""
-            } else {
-                recorder.anyValueGenerator().anyValue(retType, invocation.method.returnTypeNullable) {
-                    isTemporaryMock = true
-                    recorder.mockFactory.temporaryMock(retType)
-                }
-            }
+        val temporaryMock: () -> Any = {
+            isTemporaryMock = true
+            recorder.mockFactory.temporaryMock(retType)
+        }
+        val retValue = when {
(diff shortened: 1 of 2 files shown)


Raibaz:
Thanks a lot for taking the time to investigate this and to provide two different solutions!

In fact, when I wrote #387 and then #536 I only took into account the AnyValueGenerator usage made by the `defaultAnswer` function in `MockKStub`, as the case detailed by #565 was not covered by any tests and I didn't notice I was breaking it.

I personally like solution 1 better, because it looks more self-contained and pertaining only the case you are fixing, without impacting other clients of the `AnyValueGenerator`.

I don't think it's a big deal having to keep the collection types in sync between the RecordingState and the AnyValueGenerator, as we already cover most of the existing collection types and I don't think many more will be added in the future.

If you can please revert aeb8baa8bccaaaddab2c79a8c7c3cf08e2322fd6, or forcepush 5146a1690cbc1792a443f0783eb7df0713a31adc so that solution 1 is the one being merged, I'll merge this straight away.

Thanks again!

AlexKrupa:
Thanks for replying! I reverted the commit and squashed the other changes into one.

Issue #436, opened 2020-05-04
Real Partial Mocks and Real Partial Relaxed Mocks


ATXMJ:
Would it be possible for Mockk to support "real partial mocks" (as well as "real partial relaxed mocks", via a feature equivalent to [Mockito's `thenCallRealMethod()` feature](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#partial_mocks).

### Expected Behavior

You can create a mock that has all functions mocked **except** for a specified method, via a DSL "answer" syntax. I'd suggest something like:
```
val car = mockk<Car>(relaxed = true)
every { car.drive(any()) } calls realFunction
```

This would cause invocations of `car.drive()` to use the actual implementation.

### Current Behavior

Currently, to get this behavior, you'd need to use a `spyk<Car>()` object, and mock every function in the class **except** for the one you want to use the real implementation.

oleksiyp:
Hi.

Have you tried `mockk(relaxed=true)` + `callOriginal` in `answers`?

ATXMJ:
> Hi.
> 
> Have you tried `mockk(relaxed=true)` + `callOriginal` in `answers`?

Thank you! I think this is exactly what I was looking for! I guess this still needs to be added to the documentation?

oleksiyp:
Yes, indeed, this is not described.

ATXMJ:
@oleksiyp should we close this ticket or use it for documentation on the topic?

oleksiyp:
Lets keep it

Raibaz:
This is basically just adding proper documentation for the case described in the first issue and its solution.

derlin:
Hi there ! Would be interested in taking this one. Would a section called "Relaxed Mock calling real method" below "Relaxed mock" be ok ?

Raibaz:
Yep I think that'd be ok.

Thanks!

Raibaz:
Closed by #710.

PR #710, opened 2021-10-04
Document real partial mock

derlin:
Add a section about calling the original method. #436.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ every { func() } returns Car() // or you can return mockk() for example
 func()
 ```
 
-### Real partial (relaxed) mock
+### Partial mocking
 
 Sometimes, you need to stub some functions, but still call the real method on others, or on specific arguments.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -272,6 +272,25 @@ every { func() } returns Car() // or you can return mockk() for example
 func()
 ```
 
+### Real partial (relaxed) mock
```
Raibaz, line 275: I would just call this section "Partial mocking"

Raibaz (changes requested):

Raibaz:
Thanks for sending this through! I just have a minor concern on the title of the section.

README.md
derlin: Had the same feeling, thinking about renaming it to "*calling real method*" instead, but I also like your suggestion. Sold !

Raibaz:
Awesome, thanks a lot!

Issue #690, opened 2021-08-31
Migrate the build.gradle files to kotlin


Raibaz:
They are all currently written in groovy; I'd like to migrate them to the build.gradle.kts format to make the whole project Kotlin-based.

I think this can be done in steps, migrating one build.gradle file at a time - not sure about it though.

femialaka:
I would like to work on this one

jlengrand:
@femialaka still on it? :)

fejd:
I'd be interested in contributing to this as well.

Raibaz:
Sure, this can be worked on by multiple people at once as there are several build.gradle files that need to be migrated.

DeKaN:
@Raibaz Can I convert scripts from `gradle` directory into plugins using [composite builds](https://docs.gradle.org/current/userguide/composite_builds.html)?

jlengrand:
We start to be a lot all of a sudden XD. You can pick it up folks :).

fejd:
I'm starting with settings.gradle now. Would you prefer one commit per Gradle file? That might make it easier to collaborate if more people are interested.

jlengrand:
That'd be the best indeed, since a few folks are interested. Extra kudos if you mention which modules you are all planning to work on :).

vasanthdharmaraj:
I will work on `mockk/agent/api/build.gradle`.

geekiyer:
I will work on `agent/jvm/build.gradle`

geekiyer:
Working on `agent/common/build.gradle`

PR #711, opened 2021-10-04
Migrate settings.gradle to kts

fejd:
Initial commit for #690 that migrates settings.gradle to settings.gradle.kts.


Raibaz:
Thanks!

PR #738, opened 2021-11-08
Migrate the build.gradle files to kotlin for agent/api module

vasanthdharmaraj:
- The change is for the issue https://github.com/mockk/mockk/issues/690
- The migration to Kotlin DSL is done for the `mockk/agent/api/ module`.


Raibaz:
Thanks!

PR #739, opened 2021-11-09
Moves `MockKExtension`'s `unmockkAll` behavior to the `afterAll` callback

gmazzo:
This change is motivated by the comment I've dropped at https://github.com/mockk/mockk/issues/142#issuecomment-964500401.

`MockKExtension` is calling `unmockkAll` by default at the end of *each* test, which interferes with the `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`  configuration.

We discovered this issue after enabling `junit.extensions.autodetection.enabled=true` (by other rules, not this one) and all tests with cached mocks (per instance) started to fail because of the reset.

This PR moves the `unmockkAll`  behavior to the `afterAll` callback, and also introduces a new `@KeepMocks` (and a `mockk.junit.extension.keepmocks` property) to allow disabling just this feature while keeping the others of `MockKExtension`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,11 @@ fun calculateAddsValues1(@MockK car1: Car, @RelaxedMockK car2: Car) {
 }
 ```
 
+Finally, this extension will call `unmockkAll` in a `@AfterAll` callback, ensuring your test environment is clean after
+each test class execution.
+You can disable this behavior by adding the `@MockKExtension.KeepMocks` annotation to your class or globally by setting 
+the `mockk.junit.extension.keepmocks=true` property
+
 ### Spy
 
 Spies allow you to mix mocks and real objects.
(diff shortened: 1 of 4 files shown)


Raibaz:
This looks ok to me, thanks a lot for sending this through!

Could you also add a paragraph about unmocking and @KeepMocks in the user guide, in the "Junit5" section?

gmazzo:
> Junit5

Done!

Raibaz:
Thanks a lot! 

Not merging this right away as I want to make a release as soon as it's merged, otherwise the docs and the actual behavior of the library would not be the same.

Going to merge this ASAP though, probably some time next week.

gmazzo:
Ready to merge then? I don't have merge rights

gmazzo:
> Thanks a lot!
> 
> Not merging this right away as I want to make a release as soon as it's merged, otherwise the docs and the actual behavior of the library would not be the same.
> 
> Going to merge this ASAP though, probably some time next week.

I'm very sorry, I didn't want to rush you into this with my "Ready to merge then? I don't have merge rights". I've commented while reviewing my "pendings" without refreshing the ticket first 🤦

Raibaz:
No worries, I was ready to release anyway :)

PR #752, opened 2021-11-17
Add java 17 and kotlin 1.6.0 to CI

gustavovnicius:


Raibaz:
Thanks!

Raibaz:
Can you please rebase your branch on top of the current master?

Raibaz:
1.6.0 is not compiling because some deprecated methods, like `max`, were removed.

Can you please fix this? I don't know if the methods required by 1.6.0 are present in older versions, like 1.3.*, though: in that case, we can think about removing older versions from the build process and set a newer one as the default.

PR #763, opened 2021-12-10
JMH benchmark setup

tmszdmsk:
So this is a performance testing setup for `mockk` using JMH. See `README.md` for more details on how to run it.

I am not entirely sure it is the right place for the module (happy to move it to a better suited place)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,4 +3,4 @@ org.gradle.configureondemand=false
 org.gradle.jvmargs=-XX:MaxMetaspaceSize=768m
 localrepo=build/mockk-repo
 # localrepo=/Users/raibaz/.m2/repository
-# kotlin.version=1.5.10
+kotlin.version=1.5.31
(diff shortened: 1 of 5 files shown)


gradle.properties
```diff
@@ -3,4 +3,4 @@ org.gradle.configureondemand=false
 org.gradle.jvmargs=-XX:MaxMetaspaceSize=768m
 localrepo=build/mockk-repo
 # localrepo=/Users/raibaz/.m2/repository
-# kotlin.version=1.5.10
+kotlin.version=1.5.31
```
Raibaz, line 6: I think this should stay commented (or even better removed), because the kotlin version is being set externally by the github action that runs the builds for several different versions.

It's likely the reason why builds on 1.3.* fail.

Raibaz (changes requested):
LGTM pending a minor change.

PR #769, opened 2021-12-28
Removed unnecessary throws exception from test class

SampathKumarAmex:
`throws Exception` is not required in these test methods and flagged as warnings by IDE.


Raibaz:
Thanks!

PR #785, opened 2022-02-02
Support mocking non-accessible parent methods

### LINKED ISSUE #425: can't find private function inside superclass for dynamic call
### Expected Behavior

should find the private function inside superclass 

### Current Behavior

can't find function callPrivate() for dynamic call

#### Context

* MockK version: 1.9.3
* OS: Mac OS
* Kotlin version:  1.3.31
* JDK version: 1.8

#### Failure Logs
io.mockk.MockKException: can't find function callPrivate() for dynamic call

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
	io.mockk.MockKException: can't find function callPrivate() for dynamic call

	at io.mockk.InternalPlatformDsl.dynamicCall(InternalPlatformDsl.kt:122)
	at io.mockk.MockKMatcherScope$DynamicCall.invoke(API.kt:1969)
// -----------------------[ YOUR STACK TRACE ENDS HERE ] -----------------------
```

```kotlin
class TestPrivateCall {
  open class Parent {
        fun call() {
            callPrivate()
        }
        private fun callPrivate() {}
    }

    class Child: Parent() {}

    @Test
    fun testPrivateCall() {
        val mock = spyk(Child(), recordPrivateCalls = true)

        mock.call()

        verify { mock["callPrivate"]() }
    }
}
```

npars:
- Add support for mocking non-accessible parent methods using Mockk
- The class being mocked is always searched first for the specified
  method, then all superclasses are checked in no specific order. If
  multiple superclasses have matching signatures for the method then
  there is no guarantee which will be chosen.
- Fixes #425

### DIFF
--- a/dsl/jvm/src/main/kotlin/io/mockk/InternalPlatformDsl.kt
+++ b/dsl/jvm/src/main/kotlin/io/mockk/InternalPlatformDsl.kt
@@ -7,10 +7,12 @@ import java.lang.reflect.Method
 import java.util.concurrent.atomic.AtomicLong
 import kotlin.coroutines.Continuation
 import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
 import kotlin.reflect.KMutableProperty1
 import kotlin.reflect.KProperty1
 import kotlin.reflect.KType
 import kotlin.reflect.KTypeParameter
+import kotlin.reflect.full.allSuperclasses
 import kotlin.reflect.full.functions
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.jvm.isAccessible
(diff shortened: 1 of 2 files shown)


npars:
Regarding the issue where multiple superclasses have the same method signature. I think that could be solved in the future by allowing the user to "hint" what class contains the method they wish to mock somehow. That would guarantee that the correct method ends up getting mocked.

Raibaz:
This looks pretty cool, thanks a lot for sending this through!

I'm a bit concerned about performances, but the fact that `sequenceOf(this)` is at the beginning of the sequence of methods being looked up makes me think existing code is not going to be affected by this change.

PR #802, opened 2022-03-15
Typo

luozejiaqun:
just runs->just Runs


Raibaz:
Uhm, why?

The actual function is not uppercased, it is correct as it is now.

luozejiaqun:
Sorry, my fault. README only mentions `just Runs` in the returning unit section, I don't know `just runs` is also fine.

PR #807, opened 2022-03-20
Recognize atLeast = 0

### LINKED ISSUE #806: Recognize atLeast = 0
> Just be aware about usual MockK support pattern. 
> Tickets are checked from time to time, replied, discussed, labeled, e.t.c.
> But real fixes are applied in a month-two month period in a bunch.
> If you think this is unacceptable, go on, join the project, change the world.

### Prerequisites

```kotlin
class AtLeastTest {
    class MockCls {
        fun op(x: Int) = Unit
    }

    @Test
    fun `at least=0 should be considered`() {
        val mock = mockk<MockCls>()
        justRun { mock.op(any()) }
        mock.op(0)
        verify(atLeast = 0, atMost = 1) { mock.op(1) }
    }
}
```
should pass

### Current Behavior

doesn't pass

#### Context

* MockK version: latest(1.12.3)

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
Verification failed: call 1 of 1: MockCls(#1).op(eq(1))). Only one matching call to MockCls(#1)/op(Int) happened, but arguments are not matching:
[0]: argument: 0, matcher: eq(1), result: -

Stack trace:
                                                                                   io.mockk.impl.InternalPlatform.captureStackTrace               (InternalPlatform.kt:125)                          
                                                                                     io.mockk.impl.stub.MockKStub.handleInvocation                (MockKStub.kt:250)                                 
                                                   io.mockk.impl.instantiation.JvmMockFactoryHelper$mockHandler$1.invocation                      (JvmMockFactoryHelper.kt:23)                       
                                                                            io.mockk.proxy.jvm.advice.Interceptor.call                            (Interceptor.kt:21)   
// -----------------------[ YOUR STACK TRACE ENDS HERE ] -----------------------
```

Possibly changing default value of atLeast to 0 is needed since 
mathematicians will think `verify(atMost=3)` will be <= 3.

uyw4687:
fix #806
atLeast = 0 doesn't work

### DIFF
--- a/mockk/common/src/main/kotlin/io/mockk/impl/verify/UnorderedCallVerifier.kt
+++ b/mockk/common/src/main/kotlin/io/mockk/impl/verify/UnorderedCallVerifier.kt
@@ -112,14 +112,18 @@ open class UnorderedCallVerifier(
                         )
                     }
                 } else {
-                    VerificationResult.Failure(safeToString.exec {
-                        "$callIdxMsg. Only one matching call to ${stub.toStr()}/${matcher.method.toStr()} happened, but arguments are not matching:\n" +
+                    if (0 in min..max) {
+                        VerificationResult.OK(listOf())
+                    } else {
+                        VerificationResult.Failure(safeToString.exec {
+                            "$callIdxMsg. Only one matching call to ${stub.toStr()}/${matcher.method.toStr()} happened, but arguments are not matching:\n" +
                                 describeArgumentDifference(matcher, onlyCall) +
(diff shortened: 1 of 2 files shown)


Raibaz:
Can you please also provide a test case with the scenario you described in #806?

uyw4687:
Added `atLeastNeverAtMostOnce`!

Raibaz:
Thanks a lot!

PR #816, opened 2022-05-02
Added a JUnit 4 rule

jonapoul:
I find myself using something like this in most of my personal projects, it saves a lot (okay not that much) boilerplate of calling `MockKAnnotations.init(this)` and `unmockkAll()` in every JUnit 4 test case.

I essentially copypasted the test case from the equivalent for `MockKExtension`, let me know if that's not okay.

### DIFF
--- a/mockk/jvm/src/test/kotlin/io/mockk/junit4/MockKRuleTest.kt
+++ b/mockk/jvm/src/test/kotlin/io/mockk/junit4/MockKRuleTest.kt
@@ -6,11 +6,9 @@ import io.mockk.every
 import io.mockk.impl.annotations.MockK
 import io.mockk.impl.annotations.RelaxedMockK
 import io.mockk.impl.annotations.SpyK
-import io.mockk.junit5.MockKExtension
 import io.mockk.verify
 import org.junit.Rule
 import org.junit.jupiter.api.Test
-import org.junit.jupiter.api.extension.ExtendWith
 import kotlin.test.assertEquals
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
@@ -45,7 +43,7 @@ class MockKRuleTest {
     }
 
     @MockK
-    private lateinit var car2: Car
+    private lateinit var car: Car
 
     @RelaxedMockK
     private lateinit var relaxedCar: Car
@@ -54,7 +52,7 @@ class MockKRuleTest {
     private var carSpy = Car()
 
     @Test
-    fun injectsValidMockInMethods(@MockK car: Car) {
+    fun injectsValidMockInClass() {
         every {
             car.recordTelemetry(
                 speed = more(50),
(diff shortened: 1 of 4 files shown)


Raibaz:
This is pretty cool, thanks a lot for submitting it!

Would you mind also updating the docs [here](https://mockk.io/#junit5) to mention that this can be done with JUnit 4 as well?

Thanks again!

jonapoul:
Added a bit next to the JUnit5 section 😃

Raibaz:
Beautiful, thanks a lot!

...I guess now the only thing left to do is figuring out why the tests are failing :)

Raibaz:
I've been digging a bit into this, and looks like parameter injection is not easy to achieve with JUnit 4.

Would you be ok removing `injectsValidMockInMethods` and `injectsValidRelaxedMockInMethods` and just having `MockKRuleTest` perform `MockkAnnotations.init(this)` and `unmockkAll()`()

jonapoul:
Hopefully that should sort it.

mockk/jvm/src/test/kotlin/io/mockk/junit4/MockKRuleTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+package io.mockk.junit4
+
+import io.mockk.every
+import io.mockk.impl.annotations.MockK
+import io.mockk.impl.annotations.RelaxedMockK
+import io.mockk.impl.annotations.SpyK
+import io.mockk.verify
+import org.junit.Rule
+import org.junit.jupiter.api.Test
```
Raibaz, line 11: This should actually be `org.junit.Test`

mockk/jvm/src/test/kotlin/io/mockk/junit4/MockKRuleTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+
```
Raibaz, line 15: You need `RunWith(JUnit4::class)` here for the tests to pass

Raibaz (changes requested):
I got the build to pass on my machine, but I can't push on your branch :)

mockk/jvm/src/test/kotlin/io/mockk/junit4/MockKRuleTest.kt
jonapoul: Changed!

mockk/jvm/src/test/kotlin/io/mockk/junit4/MockKRuleTest.kt
jonapoul: Updated

Raibaz:
Thanks a lot for following up on this!

Issue #558, opened 2021-01-11
Detect unnecessary stubbings


Raibaz:
Mockito [throws an exception](https://www.baeldung.com/mockito-unnecessary-stubbing-exception) when a function is stubbed and then not actually invoked.

This is particularly useful when doing TDD, or generally when writing tests first, because it allows to stub a function, have the test fail because it was not invoked, and then make it pass by implementing the invocation.

We could add a parameter to the `mockk()` function and the `@MockK` annotation, or a configuration flag to make it global, so that it works somewhat like this:

```
class Test {
    @MockK(detectUnnecessaryStubbing = true)
    private lateinit var collaborator: Collaborator

    private lateinit var service: Service

    @BeforeEach
    fun setup() {
        MockKAnnotations.init(this) // Perhaps this could be another good place to enable the "unnecessary stubbing detection mode"

        service = Service(collaborator)
    }

    @Test
    fun test() {
        every { collaborator.function() } returns "foo"

        service.anotherFunction() // Not executing collaborator.function in here will fail
    }
}
```

A potential issue I see in this is the case when there are two tests executing `service.anotherFunction()`, one stubbing and invoking `collaborator.function()` and one not stubbing nor invoking it.

We need to find a way to detect that it's ok for the second test not to invoke `collaborator.function()` even if it was stubbed in another test.

xpepper:
Thanks @Raibaz for opening this issue.
I think this is a feature that people migrating to MockK from Mockito or jMock may expect to find here too, because it's a standard behaviour in those tools (they call it "strict mode", which is enabled by default), see for example https://www.baeldung.com/mockito-unnecessary-stubbing-exception or http://blog.mockito.org/2017/01/clean-tests-produce-clean-code-strict.html.

Raibaz:
In the end, this is basically about having [verification confirmation] executed by default. 

Perhaps it may be just about having the MockkExtension implement AfterEachCallback (see [here](https://www.baeldung.com/junit-5-extensions)) and call `confirmVerified` if necessary.

Raibaz:
Turns out this is actually a duplicate of #334.

xpepper:
@Raibaz so, is there already a solution to this issue? Thanks!

Raibaz:
Not yet, but as it turns out implementing it is likely going to be a bit easier than I thought :)

xpepper:
Great, I'm willing to test a version of MockK with this new behaviour as soon as is available!

[...]

PR #825, opened 2022-05-18
Better detection of unnecessary stubbing

PHaroZ:
[Automatic verification confirmation](https://github.com/mockk/mockk#automatic-verification-confirmation) introduces a new mechanism that ensure that all declared stubs are verified. This is cool but doesn't exactly match which is requested in #558 and exposed in [mockito doc](http://site.mockito.org/javadoc/current/index.html?org/mockito/exceptions/misusing/UnnecessaryStubbingException.html)

This PR add a new `checkUnnecessaryStub()` method, and its companion JUnit annotion `@CheckUnnecessaryStub`.

The main usecase of this feature is to avoid the `verify` steps when all stubs are done with precision and every stubs are used only once.

### DIFF
--- a/mockk/common/src/main/kotlin/io/mockk/impl/stub/MockKStub.kt
+++ b/mockk/common/src/main/kotlin/io/mockk/impl/stub/MockKStub.kt
@@ -18,6 +18,7 @@ open class MockKStub(
     val log = gatewayAccess.safeToString(Logger<MockKStub>())
 
     private val answers = InternalPlatform.synchronizedMutableList<InvocationAnswer>()
+    private val answerUsages = InternalPlatform.synchronizedMutableMap<InvocationAnswer, Int>()
     private val childs = InternalPlatform.synchronizedMutableMap<InvocationMatcher, Any>()
     private val recordedCalls = InternalPlatform.synchronizedMutableList<Invocation>()
     private val recordedCallsByMethod =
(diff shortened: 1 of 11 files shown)


mockk/common/src/main/kotlin/io/mockk/impl/stub/MockKStub.kt
```diff
@@ -32,13 +33,15 @@ open class MockKStub(
...
         answers.add(invocationAnswer)
+        answerUsages[invocationAnswer] = 0
     }
 
     override fun answer(invocation: Invocation): Any? {
         val invocationAndMatcher = InternalPlatform.synchronized(answers) {
             answers
                 .reversed()
                 .firstOrNull { it.matcher.match(invocation) }
+                ?.also { answerUsages[it] = answerUsages[it]!! + 1 }
```
Raibaz, line 44: How about using `answerUsages.getOrDefault(it, 0)` here rather than the !!?

I know `answerUsages[it]` is definitely not null because you initialize it in `addAnswer`, but `getOrDefault` seems safer to me.

Raibaz (approved):
This is pretty cool, thanks a lot!

I just have a doubt about a minor change, but apart from that LGTM.

Raibaz:
Still, I really don't know why tests are failing though.

Would you please take a look? Thanks again!

PHaroZ:
This seems due to the fact I put `InvocationAnswer` as a key of a HashMap, the hashCode of Answer try to be computed but it fails for some reason thaht I don't understand well (linked to io.mockk.impl.recording.states.CallRecordingState#cancelAndThrowBadRecordingState ).

I changed the implementation of my feature and get rid of the HashMap. All tests should be green :)

Raibaz:
Thanks again!

PR #829, opened 2022-05-24
Dependency updates, JDK17

hduerkop:


.github/workflows/gradle.yml
```diff
@@ -11,13 +11,8 @@ jobs:
     strategy:
       matrix:
         java-version: [11, 12, 13, 14, 15, 16, 17]
-        kotlin-version: [1.4.32, 1.5.31, 1.6.0]
+        kotlin-version: [1.6.21, 1.7.0-RC]
```
Raibaz, line 14: I'd keep a 1.4.* and a 1.5.* here.

Raibaz:
Looks like the `versionName` property is not available in the android gradle file:

>  /home/runner/work/mockk/mockk/mockk/android/build.gradle.kts:32:9: Unresolved reference: versionName

Can you PTAL?

Thanks!

Raibaz:
All gradle builds fail with 

> Caused by: java.lang.IllegalArgumentException: error: invalid source release: 17

I'm guessing either gradle needs to be upgraded, or JDK 17 is not supported yet.

Raibaz:
Can you please make sure this builds with earlier JDK versions and not just with 17?

Thanks

agent/android/build.gradle
```diff
@@ -59,8 +59,8 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 21
-        targetSdkVersion 31
+        minSdkVersion 26
```
AlvaroBrey, line 62: Was the minSdkVersion bump needed? 26 is oddly high for a library and will break a lot of builds.

PR #844, opened 2022-07-12
refreshVersion Plugin usage to monitor outdated dependencies

hduerkop:


Raibaz:
Can you please make sure this builds with earlier JDK versions and not just with 17?

Thanks

PR #862, opened 2022-07-25
Update 'all tests' GitHub action, enable Gradle Build Cache

aSemy:
This PR contains some proposals for improving the 'all tests' GitHub action. I hope to make them faster.

I'm making a PR before an issue to test/demo that it works as intended :)

* Only test LTS JDK versions (11, 17), plus the latest version (18 at the moment - this can change more frequently). I'm making an assumption here that there's not going to be major differences between the non-LTS versions, and even if there are, as far as I know they're not well supported. 

   Reducing the number of versions improves the actions speed, as I think there's a limit on how many jobs can run in parallel.
   
* Add 'concurrency'. I've noticed when I've pushed a few changes in rapid succession, each change will be tested. This isn't necessary - the tests should be cancelled and restarted if there's a newer commit. 
   
   I copied the concurrency config from here: https://docs.github.com/en/actions/examples/using-concurrency-expressions-and-a-test-matrix#understanding-the-example

* I bumped Kotlin 1.6.0 to 1.6.21. It seemed odd that 1.5 and 1.7 were on the latest version, but 1.6 wasn't.

* I updated the Gradle caching, copied from 

   https://github.com/JetBrains/intellij-platform-plugin-template/actions/runs/242898088/workflow#L46-L58

   This means the cache key is less specific (I think sharing the Gradle cache between Kotlin & JVM versions is fine, Gradle will rebuild/reuse tasks if necessary), and the wrapper and dependencies are cached separately (if the dependencies change, the wrapper can still be cached)

* I enabled [Gradle Build cache](https://docs.gradle.org/current/userguide/build_cache.html). This should really help build speeds.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -10,9 +10,9 @@ jobs:
   tests:
     strategy:
       matrix:
-        java-version: [11, 12, 13, 14, 15, 16, 17]
-        kotlin-version: [1.5.31, 1.6.0, 1.7.10]
-        kotlin-ir-enabled: [true, false]
+        java-version: [ 11, 17, 18 ] # test LTS versions, and the newest
+        kotlin-version: [ 1.5.31, 1.6.21, 1.7.10 ]
+        kotlin-ir-enabled: [ true, false ]
       # in case one JDK fails, we still want to see results from others
       fail-fast: false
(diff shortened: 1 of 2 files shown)


Raibaz:
This is awesome, thanks so much!

Issue #838, opened 2022-07-05
unmockkAll doesn't work if mockkConstructor was called more than once on the same class


Chrostoq:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

unmockkAll should be able to clear all mocks

### Current Behavior

Calling mockkConstructor multiple times result in unmockkAll not being able to clear it so the actual class constructor can be used again

### Failure Information (for bugs)

`internalMockkConstructor` only saves the cancellation once for each class with the `putIfNotThere`. But the `JvmConstructorMockFactory` still adds a `ConstructorMockVariant` to its stack 

#### Steps to Reproduce

1. Call mockkConstructor multiple times on the same class and overwrite a function
2. Call unmockkAll
3. Verify if the function is still mocked

#### Failure Logs

expected: <10> but was: <30>
Expected :10
Actual   :30

#### Stack trace

```
org.opentest4j.AssertionFailedError: expected: <10> but was: <30>
	at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at app//org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
	at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1152)
	at app//kotlin.test.junit5.JUnit5Asserter.assertEquals(JUnitSupport.kt:32)
	at app//kotlin.test.AssertionsKt__AssertionsKt.assertEquals(Assertions.kt:63)
	at app//kotlin.test.AssertionsKt.assertEquals(Unknown Source)
	at app//kotlin.test.AssertionsKt__AssertionsKt.assertEquals$default(Assertions.kt:62)
	at app//kotlin.test.AssertionsKt.assertEquals$default(Unknown Source)
	at app//com.example.myapplication.ExampleUnitTest.testMocking(ExampleUnitTest.kt:32)
	at java.base@11.0.12/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

```

#### Minimal reproducible code (the gist of this issue)

```
dependencies {
    testImplementation(platform('org.junit:junit-bom:5.8.2'))
    testImplementation('org.junit.jupiter:junit-jupiter')
[...]

PR #870, opened 2022-07-28
Fix unmockkAll to work if constructor was mocked multiple times

Chrostoq:
Hi all! 

This PR is to fix the Issue #838 

I changed `internalMockkConstructor` to work like `internalMockkObject` and `internalMockkStatic` by changing it to use `cancelPut`

I added a new test but I had to remove `test2`.
I'm unsure what exactly they test, but with my change the mocked instance in `ExampleObject` `private val exampleClass` is reset once mockkConstructor is called again in test2. And since it's in an object which is not reset between test1 and test2, test2 fails because it is not re-instantiated with the mocked class.

### DIFF
--- a/dsl/common/src/main/kotlin/io/mockk/API.kt
+++ b/dsl/common/src/main/kotlin/io/mockk/API.kt
@@ -526,7 +526,7 @@ object MockKDsl {
 
             MockKCancellationRegistry
                 .subRegistry(MockKCancellationRegistry.Type.CONSTRUCTOR)
-                .putIfNotThere(it, cancellation)
+                .cancelPut(it, cancellation)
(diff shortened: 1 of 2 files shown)


Raibaz:

The code LGTM, can you just add some test coverage for cases when `constructedWith` is used?

I think that may interfere with constructor mocking behavior as well and we may want to test  that `constructedWith` can be invoked several times and work correctly, both with and without using `unmockkAll` in between.

Thanks again!

Chrostoq:
Hi @Raibaz 

I updated it to include some testing for `constructedWith` as requested

Raibaz:
Thanks a lot!

PR #890, opened 2022-08-23
fix: handle isValue exceptions

qoomon:


Raibaz:
Not a big fan of catching `Throwable`.

Can we provide explicit casts for each exception type?

qoomon:
I'm not either :-), however the known exceptions, that need to be catched are `kotlin. UnsupportedOperationException`  and`kotlin.reflect.jvm.internal.KotlinReflectionInternalError`. Unfortunately `KotlinReflectionInternalError` is an internal `Error`. Technically I want to catch Exceptions and Errors, but that is Throwable in the end :-(

qoomon:
If it is okay I could catch `kotlin.reflect.jvm.internal.KotlinReflectionInternalError` and `UnsupportedOperationException` just for the jvm package and keep the other catch blocks as they are, just catching `UnsupportedOperationException`

Raibaz:
Yep this looks much better, thanks!

Issue #13, opened 2017-12-26
Doc request. Speed comparison against Mockito


stepango:
I'm using Mockito on my current project. Turns out that it's very slow. Would be great to see some benchmarks.

oleksiyp:
Why are you sure it's mockito is adding slow down?

Anyway, I'll try to benchmark it against mockito. It won't probably be very fast.

The problem here is that the speed is very dependent on use-case.
Do you have your project on GH? 
Few things are quite important:
 * how much classes do you have?
 * how much tests do you have?
 * do you have intensive computation involving mock calls?

Class re-transformation definitely adds slow down, but it's cached.
So if you have really big code-base with a lot of classes it can eat some time.
And that is the case for both frameworks

stepango:
99% sure😁 Bytecode transformations + reflection at runtime is very resource consuming. unfortunately, the project is not public. There are 3k+ tests and most of them are heavily use Mockito. As a result tests execution time ~3min on my machine, which is very long time for unit-tests.

johnny:
From my (limited) experience, mockk seems to be slow. A single unit tests with one RelaxedMock, one method stub and one verification took more than 1 second to execute with Java 8.

oleksiyp:
That may be the case. Did you have a chance to compare it to mockito? Im
interested to make it faster.

Definitely class transformation what makes it slow. If you reuse classes,
then it caches it and shouldnt be so slow for subsequent calls. Of course
something probably can be improved even now.

I understand your case is simple. But anyway if you can share your case as
github project I can exactly reproduce it and iterate on performance
optimiaztion as far as I can go

29 янв. 2018 г. 2:24 PM пользователь "Jonas von Andrian" <
notifications@github.com> написал:

oleksiyp:
No additional information provided, so no possibility to improve. Will get back to it if use case appear

ratiofu:
[...]

PR #904, opened 2022-08-25
introduce Kotlinx Benchmark tests (copied from #763)

aSemy:
copied from #763 - credit to @tmszdmsk!

I thought this was a good change, and I'm curious about investigating #13 and I think these benchmark tests can help.

I also refactored the existing 'client tests' to be in a new 'test modules' directory, because I think it's neater and more organised to have the test projects together, rather than mixed amongst the 'regular' projects.


test-modules/performance-tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+to build the `.jar` file containing the benchmark code. It will be saved to `./build/benchmarks/main/jars` directory.
+
+From there you can run JMH with all the JMH-specific configuration options form CLI (even on different machine).
+
+```shell
+java -jar ./build/benchmarks/main/jars/performance-tests-main-jmh-<version>.jar -w 1 -i 2 -r 60s -prof jfr
+```
+
+which will attach the JFR profiler (not possible with `kontlinx-benchmark` executed from Gradle)
```
Raibaz, line 31: Typo in `kontlinx-benchmark` :)

Raibaz (commented):
This looks pretty interesting, thanks for putting it together!

aSemy:
I added some additional tests for context, and used https://jmh.morethan.io/ to visualise the results. 

image


I don't think it's a surprise that the test without mocking/stubbing is very fast, but as commented [here](https://github.com/mockk/mockk/issues/13#issuecomment-990576933) the stub decreases the operations, even though the stub isn't called.

Raibaz:
Yep makes sense, as discussed in #13 the most expensive part is actually creating stubs rather than using them.

aSemy:
I think this PR can be merged in as-is. It's a good basis for future work, be it making the code better, or adding a GHA or visualizer plugin https://github.com/jzillmann/gradle-jmh-report

Raibaz:
Yep I think the next steps about this would be adding a GHA to monitor performances and track potential hits.

Thanks a lot!

PR #912, opened 2022-08-29
GHA Snapshot publishing

### LINKED ISSUE #910: Automatically release the latest version to Sonatype snapshot
Rather than asking people to manually checkout the project and release to Maven Local, we should set up a GitHub Action to automatically deploy every successful build (on the master branch) to the Sonatype Snapshot repo.

This would help with getting earlier feedback and preventing issues during release.

This would require

* a new GitHub Action - easy
* Minor Gradle config update - again, easy
* Setting the tokens in GitHub project secrets, and exposing them to the GHA

aSemy:
Publish a snapshot version to Sonatype on every successful build on master branch.

Fixes #910 

This requires setting GitHub secrets  - see `.github/workflows/publish.yml`


Raibaz:
Thanks!

A couple of notes: according to [this doc](https://docs.github.com/en/actions/security-guides/encrypted-secrets), i guess you're right and periods need to be removed.

Also, apparently I cannot set the secrets in the repo, so I'm afraid we need to summon @oleksiyp

aSemy:
The Gradle secrets `${{ secrets.GRADLE_SIGNING_KEY_ID }}` can't have periods, but what about the environment variables they're assigned to? `ORG_GRADLE_PROJECT_signing.key`? For simplicity I'll rename the environment variables to match.

luongvo:
@Raibaz Could you please resume reviewing this soon? 👋

Raibaz (approved):

Raibaz:
PR LGTM, but I can't set the secrets :(

oleksiyp:
Do you need my help?

buildSrc/src/main/kotlin/buildsrc/convention/mockk-publishing.gradle.kts
```diff
@@ -47,21 +41,35 @@ val mavenName: String by project.extra
...
+            maven("https://oss.sonatype.org/content/repositories/snapshots/") {
+                name = "SonartypeStaging"
+                credentials {
+                    username = ossrhUsername.get()
+                    password = ossrhPassword.get()
+                }
+            }
+
+            if (!isSnapshotVersion.get()) {
+                maven("https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
```
oleksiyp, line 62: Please check this guide https://central.sonatype.org/publish/publish-gradle/.
I have doubts that this is right URL.
I will try to deal with required secrets and set them in GH settings:
 - secrets.GRADLE_SIGNING_KEY_ID
 - secrets.GRADLE_SIGNING_PASSWORD
 - secrets.GRADLE_SIGNING_SECRET_KEY_RING_FILE
 - secrets.OSS_SONATYPE_USERNAME
 - secrets.OSS_SONATYPE_PASSWORD

aSemy: Thanks @oleksiyp!

Would you consider [granting @Raibaz ownership](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-peoples-access-to-your-organization-with-roles/maintaining-ownership-continuity-for-your-organization)? Then he could also manage the secrets.

> I have doubts that this is right URL.


MockK was published before 2021 so still uses the 'legacy host', but the docs mostly use the new URLs.

> Note: As of February 2021, all new projects began being provisioned on https://s01.oss.sonatype.org/. If your project is not provisioned on https://s01.oss.sonatype.org/, you will want to reference the legacy host https://oss.sonatype.org/.

The current MockK config uses the same URL and it works okay

https://github.com/mockk/mockk/blob/1419517e0de42c6148d0fb325467d16797ed3e53/buildSrc/src/main/kotlin/buildsrc/convention/mockk-publishing.gradle.kts#L15-L21

oleksiyp: Not sure you will need other secrets... and not sure you need them for something else... security is always compromise of comfort. So no 😄  but remember that I am still alive and happy to help 😄

oleksiyp: Afterall if @Raibaz feels that more access is required happy to understand the reasoning and cooperate.

There is a bigger issue where to spend $4k on https://opencollective.com/mockk. Previously I thought it was too small an amount to spend paying contributors... but now it is not that obvious

oleksiyp:
@Raibaz @aSemy 

I added a few secrets
image

***Please note there is no way to delimit release access and snapshot access... so this is my creds. I ask you to be careful with build scripts PR review so that nobody does "echo $KEY" stuff. I believe it should be protected somehow, but I never tried to use such a hack in GH actions. In Jenkins such a hack is easy peasy, to be honest. ***

Also, @aSemy, note that local run suppose you have a file with keyring, while I cannot import binary files into GH action secret field.
So you will need to adopt "import to gpg" somehow
```
- name: Configure GPG Key
  run: |
    echo -n "$GPG_SIGNING_KEY" | base64 --decode | gpg --import
  env:
    GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
```
Especially for that purpose name secret is a "signing key", and not a "signing key ring file".
You may count that the private signing key unencrypted in PEM format is stored there.
The same reason "password" is not set, coz anyway you would need first to import to gpg.

Also, note I am going to go to Paris tomorrow and will not have access to the laptop where the original key is stored. GH does not allow to read values as well, so hope what I set is enough.

Feel free to ask any questions.

aSemy:
thanks @oleksiyp, I will take a look! I think there's a way of 'mounting' a GitHub Secret as a file...

There are a number of protections for GitHub Secrets, but there's [a couple more](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) that could be done

* [prevent PRs from creating/approving PRs](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#preventing-github-actions-from-creating-or-approving-pull-requests)
* Require that any PR that edits a GitHub Workflow must be [reviewed and approved by `CODEOWNERS`](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-codeowners-to-monitor-changes)

These require configuration at the project and/or org level, so it's up to you to implement them @oleksiyp :)

buildSrc/src/main/kotlin/buildsrc/convention/mockk-publishing.gradle.kts
Raibaz: I'm perfectly fine with the access I have now, I don't think I need more :)

After all, this should be a one-shot operation and we should be ok with secrets now, I don't think I'll need access to change them anytime soon.

Raibaz:
Thanks a lot @oleksiyp!

I think once we put in place the two protections that @aSemy suggested we should be fine.

aSemy:
@Raibaz Thanks for the reminder, I'll pick this up again, hopefully by the end of the week. Since I last looked at this [I set up automatic publishing for one of my libraries](https://github.com/adamko-dev/kotka-streams/compare/99467d0f869ea89989730c75c3c6f3b7be609de8...263a80d8a0eb4b5ec4cb867cb70b808f2f8d346e), so I've got a better idea of how things work now.

aSemy:
@Raibaz as far as I can tell, this PR should be ready to go. Because I don't have access to the secrets or Sonatype Nexus I can't tell if it will work though, and I also won't be able to debug it. I think it's best if you or @oleksiyp pick it up, since you'll be able to more easily see what's going on. It might take some more tweaking. If there's any problems, then let me know and I can try and help out.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: adopt
+          java-version: 11
+
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
```
TWiStErRob, line 30: there are some new releases:

```suggestion
    steps:
      - uses: actions/checkout@v4

      - uses: actions/setup-java@v4
        with:
          distribution: adopt
          java-version: 11

      - name: Setup Gradle
        uses: gradle/actions/setup-gradle@v4
```

no further changes required for these majors.

esafak:
@Raibaz Are we good here?

PR #913, opened 2022-09-04
De-duplicate `ValueClassSupport`

### LINKED ISSUE #857: Reduce code duplication across subprojects
There is some amount of code duplication in the MockK modules. Specifically `ValueClassSupport` (but there might be more?)

https://github.com/mockk/mockk/blob/54e30ec66e4aeb61942b976dbb33fd37c9487629/agent/android/src/main/kotlin/io/mockk/ValueClassSupport.kt
https://github.com/mockk/mockk/blob/54e30ec66e4aeb61942b976dbb33fd37c9487629/agent/jvm/src/main/kotlin/io/mockk/ValueClassSupport.kt

Eliminating this code duplication would reduce the risk of bugs and inconsistent behaviour.

### See also 

* This issue was made because of #849, discussion here https://github.com/mockk/mockk/pull/849#pullrequestreview-1048197048
* https://github.com/mockk/mockk/issues/828
* https://github.com/mockk/mockk/pull/653

aSemy:
* Create new module for platform implementations (there was no existing suitable project, due to the internal dependencies between them), 
* move `ValueClassSupport` into it (deleting the duplicated versions in Android and DSL modules)

Fixes #857 

Are there any other classes that can be de-duplicated?


aSemy:
@qoomon Can you take a look? What do you think?

Raibaz:
LGTM, waiting for a check by @qoomon as well.

Thanks!

qoomon:
Sorry for the late response, I'll do the review now.

qoomon (approved):
LGTM, however I think the wording platform is a little misleading.

Is there a reason why don't implement `ValueClassSupport`within the common mockk module?
https://github.com/aSemy/mockk/tree/fix/de-duplicate-valueclasssupport/modules/mockk/src/commonMain/kotlin/io/mockk

aSemy:
> LGTM, however I think the wording platform is a little misleading.

Yeah I think the name could be better. `mockk-core`, `mockk-common`, `mockk-shared-tools`, `mockk-tools`?

> Is there a reason why don't implement `ValueClassSupport`within the common mockk module? https://github.com/aSemy/mockk/tree/fix/de-duplicate-valueclasssupport/modules/mockk/src/commonMain/kotlin/io/mockk


Because `mockk` depends on `mockk-android` and `mockk-dsl`

https://github.com/aSemy/mockk/blob/df2ccd4f3f8575554c3083b957bc3b0e36827c98/modules/mockk/build.gradle.kts#L22-L25

it can't be depended on by the other modules. 

If I were to refactor the project then I would have the root project be `io.mockk:mockk`, and it would depend on the subprojects as API dependencies and not have *any* code. Then I'd rename `io.mockk:mockk` to `io.mockk:mockk-core`, and that would be the shared place for code. `mockk-android` would be an extension of `mockk-core`.

qoomon:
Thanks for the explanation. Regarding the module name I'd vote for mockk-core.

Raibaz:
+1 for mockk-core.

aSemy:
Okay, renamed! I changed the package too.

I also removed `fun unboxClass(cls: KClass<*>): KClass<*>` from `InternalPlatformDsl`, since it was just a re-name of the `ValueClassSupport,boxedClass`  util.

Raibaz:
LGTM

@qoomon ?

qoomon:
Love it.

qoomon:
@aSemy thanks for your effort

PR #914, opened 2022-09-04
use the official Gradle GHA

aSemy:
Use the official Gradle GitHub Action https://github.com/marketplace/actions/gradle-build-action

Two reasons: 

1. I've noticed that the GitHub actions aren't always cached, and thus they take longer than necessary
2. Less configuration is needed

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -83,4 +83,7 @@ jobs:
           script: echo "Generated AVD snapshot for caching."
 
       - name: run tests
-        run: ./gradlew connectedCheck
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: ${{ matrix.api-level }}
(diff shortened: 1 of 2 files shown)


.github/workflows/gradle.yml
```diff
@@ -24,31 +24,22 @@ jobs:
...
         with:
-          arguments: |
-            test 
-            --stacktrace 
-            -Pkotlin.version=${{ matrix.kotlin-version }} 
+          cache-read-only: false
+
+      - name: Run tests with Gradle
+        run: |
+          ./gradlew check
```
aSemy, line 40: I also changed `./gradlew test` to `./gradlew check`, because I noticed that the API dumps are not up to date. The `apiCheck` task will run if the `check` task runs, not `test.

Raibaz:
Thanks a lot, especially for spotting that we were doing `gradle test` rather than `gradle check`!

PR #915, opened 2022-09-04
check for sealed classes in ObjenesisInstantiator

### LINKED ISSUE #832: java.lang.InstantiationError when upgrading to Kotlin 1.7
Java 17 + Kotlin 1.7 + Mockk 1.12.4/Spring Mockk 3.1.1

```bash 
java.lang.InstantiationError: com.example.demo.UpdateAccountResult
    at jdk.internal.reflect.GeneratedSerializationConstructorAccessor18.newInstance(Unknown Source)
    at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
    at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
    at org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator.newInstance(SunReflectionFactoryInstantiator.java:48)
```

It occurred whens using `coEvery{}` to do some stub, which returns a `UpdateAccountResult.Success()`.

Here `UpdateAccountResult` is a `sealed class`, it has some detailed result subclass/sub data classes.

And `UpdateAccountResult` extends from a base abstract class like this.

```kotlin 
abstract class ApiBaseResult {
    @field:JsonProperty("SC")
    val statusCode: String? = null

    @field:JsonProperty("EC")
    val errorCode: String? = null

    @field:JsonProperty("EM")
    val errorMessage: String? = null
}
``` 

These codes are working well with **Kotlin 1.6.21**.

aSemy:
Trying to fix #832

### DIFF
--- a/modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ObjenesisInstantiator.kt
+++ b/modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ObjenesisInstantiator.kt
@@ -24,6 +24,10 @@ class ObjenesisInstantiator(
         if (cls == Any::class.java) {
             @Suppress("UNCHECKED_CAST")
             return Any() as T
+        } else if (cls.isSealed) {
+            cls.permittedSubclasses.firstNotNullOfOrNull { subCls ->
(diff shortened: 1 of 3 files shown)


aSemy:
Continued in https://github.com/mockk/mockk/pull/916

PR #916, opened 2022-09-06
Properly handle sealed classes with Kotlin 1.7 and JDK 17

stuebingerb:
Attempt to fix https://github.com/mockk/mockk/issues/832 by supporting sealed classes based on previous work done by @aSemy.

### DIFF
--- a/modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ProxyMaker.kt
+++ b/modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ProxyMaker.kt
@@ -105,6 +105,12 @@ internal class ProxyMaker(
         } else if (interfaces.isEmpty() && !Modifier.isAbstract(clazz.modifiers) && inliner != null) {
             log.trace("Taking instance of $clazz itself because it is not abstract and no additional interfaces specified.")
             clazz
+        } else if (clazz.kotlin.isSealed) {
+            log.trace("Taking instance of subclass of $clazz because it is sealed.")
+            clazz.kotlin.sealedSubclasses.firstNotNullOfOrNull {
+                @Suppress("UNCHECKED_CAST")
+                subclass(it.java, interfaces) as Class<T>
+            } ?: error("Unable to create proxy for sealed class $clazz, available subclasses: ${clazz.kotlin.sealedSubclasses}")
         } else {
             log.trace(
                 "Building subclass proxy for $clazz with " +
@@ -143,7 +149,7 @@ internal class ProxyMaker(
             val defaultConstructor = cls.getDeclaredConstructor()
             try {
                 defaultConstructor.isAccessible = true
-            } catch (ex: Exception) {
+            } catch (ignored: Exception) {
                 // skip
             }
 
--- a/modules/mockk/src/commonTest/kotlin/io/mockk/it/SealedClassTest.kt
+++ b/modules/mockk/src/commonTest/kotlin/io/mockk/it/SealedClassTest.kt
@@ -2,7 +2,6 @@ package io.mockk.it
 
 import io.mockk.every
 import io.mockk.mockk
-import kotlin.test.Ignore
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
@@ -10,7 +9,6 @@ import kotlin.test.assertEquals
 class SealedClassTest {
 
     @Test
-    @Ignore("Fails on JDK17+ https://github.com/mockk/mockk/issues/832")
     fun serviceReturnsSealedClassImpl() {
         val factory = mockk<Factory> {
             every { create() } returns Leaf(1)
@@ -22,7 +20,6 @@ class SealedClassTest {
     }
 
     @Test
-    @Ignore("Fails on JDK17+ https://github.com/mockk/mockk/issues/832")
     fun serviceAnswersSealedClassImpl() {
         val factory = mockk<Factory> {
             every { create() } answers { Leaf(1) }
--- a/modules/mockk/src/commonTest/kotlin/io/mockk/it/SealedInterfaceTest.kt
+++ b/modules/mockk/src/commonTest/kotlin/io/mockk/it/SealedInterfaceTest.kt
@@ -3,14 +3,12 @@ package io.mockk.it
 import io.mockk.every
 import io.mockk.mockk
 import kotlin.test.Test
-import kotlin.test.Ignore
 import kotlin.test.assertEquals
 
 
 class SealedInterfaceTest {
 
     @Test
-    @Ignore("Fails on JDK17+ https://github.com/mockk/mockk/issues/832")
     fun serviceReturnsSealedClassImpl() {
         val factory = mockk<Factory> {
             every { create() } returns Leaf(1)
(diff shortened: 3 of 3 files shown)


modules/mockk-agent-api/src/commonMain/kotlin/io/mockk/proxy/MockKInstantiatior.kt
```diff
@@ -1,5 +1,5 @@
 package io.mockk.proxy
 
 interface MockKInstantiatior {
-    fun <T> instance(cls: Class<T>): T
+    fun <T : Any> instance(cls: Class<T>): T
```
stuebingerb, line 4: The `.kotlin` extension is only defined for `KClass<T : Any>` but from what I've seen it had been `T : Any` in all calling sites anyway.

stuebingerb: Alright, seems I've missed something. Will revise.

stuebingerb:
Currently it seems to _not_ solve our issues, though. We have the following class hierarchy (simplified):
```
sealed class User
abstract class KnownUser : User
class UserA : KnownUser
class UserB : KnownUser
```
and `getPermittedSubclasses` returns only the `KnownUser` class (which of course cannot be instantiated either).

stuebingerb:
Currently looking into a potential open issue: although the mockk tests succeed, our own actual tests fail.

aSemy:
Would you be able to add a test using the example that failed? (A sealed class with an abstract child?)

stuebingerb:
> Would you be able to add a test using the example that failed? (A sealed class with an abstract child?)

Tried, but that test also succeeded without my modifications. I'm still missing something.

aSemy:
Yes the stacktrace helps, thanks, although I can't figure out exactly what's wrong. At least I think it shows that we're on the right path! I don't see any reference to `ObjenesisInstantiator.kt`. The last MockK code called is in `io.mockk.proxy.jvm.transformation.SubclassInstrumentation`, so something in there (or earlier) isn't handling sealed subclasses properly. I suspect a similar fix will be needed.

stuebingerb:
> Yes the stacktrace helps, thanks, although I can't figure out exactly what's wrong. At least I think it shows that we're on the right path! I don't see any reference to `ObjenesisInstantiator.kt`. The last MockK code called is in `io.mockk.proxy.jvm.transformation.SubclassInstrumentation`, so something in there (or earlier) isn't handling sealed subclasses properly. I suspect a similar fix will be needed.

My second commit here fixes the issue! I'm just wondering why and why I cannot come up with a dedicated test. Sorry for any confusion caused.

modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ObjenesisInstantiator.kt
```diff
@@ -24,6 +24,11 @@ class ObjenesisInstantiator(
         if (cls == Any::class.java) {
             @Suppress("UNCHECKED_CAST")
             return Any() as T
+        } else if (cls.isSealedSafe()) {
+            @Suppress("UNCHECKED_CAST")
+            return cls.getPermittedSubclassesSafe().firstNotNullOfOrNull { subCls ->
```
stuebingerb, line 29: @aSemy: Your PR did not have the `return` here and IIRC the mockk tests were also running successfully without. Was that an oversight or am I missing something?

aSemy: Good spot! Yes that's a mistake, I intended to use a `return`. However, it does show that Objenesis can probably handle sealed classes and we don't need to have custom code.

```kotlin
    private val instantiators = Collections.synchronizedMap(WeakHashMap<Class<*>, ObjectInstantiator<*>>())
    override fun <T> instance(cls: Class<T>): T {
        if (cls == Any::class.java) {
            @Suppress("UNCHECKED_CAST")
            return Any() as T
        } else if (!Modifier.isFinal(cls.modifiers) && !cls.isSealedSafe()) {
            try {
                val instance = instantiateViaProxy(cls)
                if (instance != null) {
                    return instance
                }
            } catch (ex: Exception) {
                log.trace(
                    ex, "Failed to instantiate via proxy " + cls + ". " +
                            "Doing objenesis instantiation"
                )
            }
        }
        return instanceViaObjenesis(cls)
    }
```

stuebingerb: Thought so as well - and the mockk tests are indeed running successfully. But in our code that leads to the original `InstantiationError`:

```
java.lang.InstantiationError: org.example.SealedClass
        at jdk.internal.reflect.GeneratedSerializationConstructorAccessor7.newInstance(Unknown Source)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:483)
        at org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator.newInstance(SunReflectionFactoryInstantiator.java:48)
        at io.mockk.proxy.jvm.ObjenesisInstantiator.instanceViaObjenesis(ObjenesisInstantiator.kt:83)
        at io.mockk.proxy.jvm.ObjenesisInstantiator.instance(ObjenesisInstantiator.kt:42)
        at io.mockk.impl.instantiation.JvmInstantiator$instantiate$2.invoke(JvmInstantiator.kt:16)
```
I don't yet have a theory _why_ but I can see that our code is coming from the subclass creation.

But it looks like actually the changes in `ProxyMaker` are enough to get our tests running. I am still struggling to create a corresponding test case in mockk, though. Will update this PR later.

stuebingerb:
Sorry for commit spam. The current implementation here now seems to fix both, mockk's tests and our own.

stuebingerb:
So... I tried to add a new test case based on our code and I'm claiming that the CI tests are currently broken. Have a look at https://github.com/stuebingerb/mockk/pull/10 where the `SealedClassTest` and `SealedInterfaceTest` are both running successfully without any adjustments to the implementation.

@aSemy Any idea what might be wrong there?

aSemy:
That is indeed weird! Good catch. I think the new Gradle GHA is re-using the cache for different JDK versions. I see this message in the logs: `Task :modules:mockk:test SKIPPED`

stuebingerb:
@Raibaz @aSemy What is needed to get this merged?

aSemy:
I thought that because there was an issue with the Gradle tests not failing https://github.com/mockk/mockk/pull/916#issuecomment-1240443435 that it wasn't clear if this PR would work as expected.

I don't have any suggestions on how to move forward...

stuebingerb:
I guess someone will have to fix the CI. But I currently cannot.

aSemy:
Let's do it then 👍

aSemy (approved):

Raibaz:
Ok makes sense, sorry I lost track of what was going on here :)

DHosseiny:
This change breaks mocking a sealed class with kotlin 1.6

PR #926, opened 2022-09-23
update dependencies in readme to reflect new multiplatform structure

aSemy:
* Removed `-jvm` dependency from Gradle dependencies - Gradle will automatically determine the correct platform
* I removed the 'common' dependency because `mockk-common` doesn't exist any more
* Changed `{version}` to `{mockkVersion}`, because in Gradle `$version` will refer to the Project version.


Raibaz:
Thanks!

Issue #934, opened 2022-09-29
Mockk cannot handle sealed classes in `every` block


ThanksForAllTheFish:
Sample project: https://github.com/ThanksForAllTheFish/mockk832
Related: https://github.com/mockk/mockk/issues/832

Mockk 1.13.2 still throws

```
org.t4atf.mockk832.Node
java.lang.InstantiationError: org.t4atf.mockk832.Node
	at jdk.internal.reflect.GeneratedSerializationConstructorAccessor17.newInstance(Unknown Source)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator.newInstance(SunReflectionFactoryInstantiator.java:48)
	at io.mockk.proxy.jvm.ObjenesisInstantiator.instanceViaObjenesis(ObjenesisInstantiator.kt:75)
	at io.mockk.proxy.jvm.ObjenesisInstantiator.instance(ObjenesisInstantiator.kt:42)
	at io.mockk.impl.instantiation.JvmInstantiator$instantiate$2.invoke(JvmInstantiator.kt:16)
	at io.mockk.impl.instantiation.AbstractInstantiator.instantiateViaInstanceFactoryRegistry(AbstractInstantiator.kt:17)
```

when using sealed classes on kotlin 1.7 and java 17. The issue now is in `every` blocks, for instance:

```
class Root : Node()
class Leaf : Node()

interface Factory {
    fun create() : Node
    fun copy(node: Node) : Node
}

open class MyTest {
    @Test
    open fun aTest() {
        val mockk = mockk<Factory>()
        every { mockk.create() } returns Leaf() // this was fixed in mockk832

        println(mockk.create())

        every { mockk.copy(any()) } returns Root() //but this still breaks
    }
}
```

aSemy:
Looks like a good demo - can you (or someone else) update https://github.com/mockk/mockk/blob/6d5fe10cfd3662c7716f7381ecc6b06bde12a091/modules/mockk/src/commonTest/kotlin/io/mockk/it/SealedClassTest.kt and make a PR?

diastremskii:
The test can be something like that:
<details>
<summary>Example test</summary>

```kotlin
class SealedInterfaceTest {

    @Test
    fun serviceTakesSealedInterfaceAsInput() {
[...]

PR #936, opened 2022-09-29
Fix instantiation error when mocking sealed classes in JDK 17

odin-delrio:
Related to #934 -

## First step, reproducing the issue:
Demonstrate the failure when using the `any()` matcher.

WIP


odin-delrio:
couldn't reproduce properly, closing and I'll give it a try later again.

PR #946, opened 2022-10-10
Add missing test annotation to LambdaTest

bossm0n5t3r:
There is no test annotation in LambdaTest. Add missing test annotation to LambdaTest.

I don't know history about this, Is there a reason why there is no test annotation?


Raibaz:
Definitely a mistake, thanks for spotting this!

PR #951, opened 2022-10-21
Fix verifier logic for slots and different matchers

### LINKED ISSUE #774: Regression: capturing subtypes in verify block fails in v1.12.1
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

### Expected Behavior

The code below works correctly in v1.9.3, capturing the single call argument with type `Impl2` into the slot.

### Current Behavior

After upgrading from v1.9.3 to v1.12.1, the code started throwing an exception with the stack trace below.
If I try to capture in a `mutableListOf<Impl2>()`, the verification succeeds, but the resulting list contains 2 elements: one of type Impl1 and one of type Impl2.

#### Steps to Reproduce

* Mock a method that can receive arguments of different types
* Verify that the call with one specific type happened only once, and capture the argument in a slot

See code below

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.12.1
* OS: macOS 
* Kotlin version: 1.6.10
* JDK version: 11
* JUnit version: 5.8.2
* Type of test: unit test

#### Stack trace

```
Service(#1).method(slotCapture<Impl2>())) execution is being verified more than once and its arguments are being captured with a slot.
This will store only the argument of the last invocation in the slot.
If you want to store all the arguments, use a mutableList to capture arguments.
io.mockk.MockKException: Service(#1).method(slotCapture<Impl2>())) execution is being verified more than once and its arguments are being captured with a slot.
This will store only the argument of the last invocation in the slot.
If you want to store all the arguments, use a mutableList to capture arguments.
	at app//io.mockk.impl.verify.UnorderedCallVerifier.matchCall(UnorderedCallVerifier.kt:62)
	at app//io.mockk.impl.verify.UnorderedCallVerifier.verify(UnorderedCallVerifier.kt:38)
	at app//io.mockk.impl.recording.states.VerifyingState$recordingDone$outcome$1.invoke(VerifyingState.kt:31)
	at app//io.mockk.impl.recording.states.VerifyingState$recordingDone$outcome$1.invoke(VerifyingState.kt:11)
	at app//io.mockk.impl.recording.CommonCallRecorder.safeExec(CommonCallRecorder.kt:72)
	at app//io.mockk.impl.recording.states.VerifyingState.recordingDone(VerifyingState.kt:30)
	at app//io.mockk.impl.recording.CommonCallRecorder.done(CommonCallRecorder.kt:47)
	at app//io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:64)
	at app//io.mockk.impl.eval.VerifyBlockEvaluator.verify(VerifyBlockEvaluator.kt:30)
	at app//io.mockk.MockKDsl.internalVerify(API.kt:118)
	at app//io.mockk.MockKKt.verify(MockK.kt:149)
	at app//io.mockk.MockKKt.verify$default(MockK.kt:146)
	at app//MockkBugTest.test(MockkBugTest.kt:33)
```

```kotlin
// -----------------------[ GRADLE DEFINITIONS ] -----------------------
dependencies {
    testImplementation("io.mockk:mockk:1.12.1")
    testImplementation("org.junit.jupiter:junit-jupiter:5.8.2")
}
// -----------------------[ YOUR CODE STARTS HERE ] -----------------------
import io.mockk.every
import io.mockk.just
import io.mockk.mockk
import io.mockk.runs
import io.mockk.slot
import io.mockk.verify
import org.junit.jupiter.api.Test

interface Interface

class Impl1 : Interface

class Impl2 : Interface

interface Service {
    fun method(obj: Interface)
}

class MockkBugTest {

    @Test
    fun test() {
        val service = mockk<Service> {
            every { method(any()) } just runs
        }

        service.method(Impl1())
        service.method(Impl2())
        
        verify(exactly = 1) { service.method(ofType<Impl2>()) } // works OK

        val slot = slot<Impl2>()
        verify(exactly = 1) { service.method(capture(slot)) } // throws
    }
}
```

m-burst:
Closes #774

### DIFF
--- a/modules/mockk/src/commonMain/kotlin/io/mockk/impl/verify/UnorderedCallVerifier.kt
+++ b/modules/mockk/src/commonMain/kotlin/io/mockk/impl/verify/UnorderedCallVerifier.kt
@@ -55,7 +55,9 @@ open class UnorderedCallVerifier(
         val allCallsForMock = stub.allRecordedCalls()
         val allCallsForMockMethod = stub.allRecordedCalls(matcher.method)
 
-        if(allCallsForMockMethod.size > 1 && matcher.args.any { it is CapturingSlotMatcher<*> }) {
+        val matchedCalls = allCallsForMockMethod.filter(matcher::match)
+
+        if(matchedCalls.size > 1 && matcher.args.any { it is CapturingSlotMatcher<*> }) {
             val msg = "$matcher execution is being verified more than once and its arguments are being captured with a slot.\n" +
                 "This will store only the argument of the last invocation in the slot.\n" +
                 "If you want to store all the arguments, use a mutableList to capture arguments."
@@ -63,7 +65,7 @@ open class UnorderedCallVerifier(
         }
 
         val result = if (min == 0 && max == 0) {
-            if (!allCallsForMockMethod.any(matcher::match)) {
+            if (matchedCalls.isEmpty()) {
                 VerificationResult.OK(listOf())
             } else {
                 VerificationResult.Failure(
@@ -127,7 +129,6 @@ open class UnorderedCallVerifier(
                 }
             }
             else -> {
-                val matchedCalls = allCallsForMockMethod.filter(matcher::match)
                 val n = matchedCalls.count()
                 if (n in min..max) {
                     VerificationResult.OK(matchedCalls)
@@ -162,7 +163,7 @@ open class UnorderedCallVerifier(
         }
 
         captureBlocks.add {
-            for (call in allCallsForMockMethod) {
+            for (call in matchedCalls) {
                 matcher.captureAnswer(call)
             }
         }
--- a/modules/mockk/src/commonTest/kotlin/io/mockk/it/CapturingTest.kt
+++ b/modules/mockk/src/commonTest/kotlin/io/mockk/it/CapturingTest.kt
@@ -85,8 +85,8 @@ class CapturingTest {
 
         assertFailsWith<MockKException> {
             verify {
-                mock.doSomething("1", capture(dataSlotId1))
-                mock.doSomething("2", capture(dataSlotId2))
+                mock.doSomething(any(), capture(dataSlotId1))
+                mock.doSomething(any(), capture(dataSlotId2))
             }
         }
     }
@@ -136,6 +136,8 @@ class CapturingTest {
             mock.doSomething("2", capture(slotList))
         }
 
+        // Each capture should have happened once because of different matchers for `id` argument
+        assertEquals(slotList.size, 2)
         assertEquals("data1", slotList[0])
         assertEquals("data2", slotList[1])
     }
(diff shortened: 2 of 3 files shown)


modules/mockk/src/commonMain/kotlin/io/mockk/impl/verify/UnorderedCallVerifier.kt
```diff
@@ -55,15 +55,17 @@ open class UnorderedCallVerifier(
         val allCallsForMock = stub.allRecordedCalls()
         val allCallsForMockMethod = stub.allRecordedCalls(matcher.method)
 
-        if(allCallsForMockMethod.size > 1 && matcher.args.any { it is CapturingSlotMatcher<*> }) {
+        val matchedCalls = allCallsForMockMethod.filter(matcher::match)
+
+        if(matchedCalls.size > 1 && matcher.args.any { it is CapturingSlotMatcher<*> }) {
```
m-burst, line 60: It's OK if there are multiple calls overall but only one of them matches and will be captured

modules/mockk/src/commonMain/kotlin/io/mockk/impl/verify/UnorderedCallVerifier.kt
```diff
@@ -162,7 +163,7 @@ open class UnorderedCallVerifier(
         }
 
         captureBlocks.add {
-            for (call in allCallsForMockMethod) {
+            for (call in matchedCalls) {
```
m-burst, line 166: `captureAnswer` was called for non-matched calls, which results in incorrect capturing results

modules/mockk/src/commonTest/kotlin/io/mockk/it/CapturingTest.kt
```diff
@@ -85,8 +85,8 @@ class CapturingTest {
 
         assertFailsWith<MockKException> {
             verify {
-                mock.doSomething("1", capture(dataSlotId1))
-                mock.doSomething("2", capture(dataSlotId2))
```
m-burst, line 89: This would not fail if left as is, because with different values of the first argument only 1 call will be matched in each case

modules/mockk/src/commonTest/kotlin/io/mockk/it/CapturingTest.kt
```diff
@@ -136,6 +136,8 @@ class CapturingTest {
             mock.doSomething("2", capture(slotList))
         }
 
+        // Each capture should have happened once because of different matchers for `id` argument
+        assertEquals(slotList.size, 2)
```
m-burst, line 140: Previously this was 4

Raibaz:
Brilliant, thanks a lot for fixing this!

Issue #929, opened 2022-09-26
mocking static on Java 19 is broken


uthark:
> Just be aware about usual MockK support pattern. 
> Tickets are checked from time to time, replied, discussed, labeled, e.t.c.
> But real fixes are applied in a month-two month period in a bunch.
> If you think this is unacceptable, go on, join the project, change the world.

## Please remove sections wisely

Below information is actually needed to make all the process of fixing faster.
Choose main points. Don't blindly follow this as a set of rules. 
Don't waste much time. Usually, the main thing is to have a good reproducible minimal code.

### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [ ] I checked the documentation and found no answer
- [ ] I checked to make sure that this issue has not already been filed

### Expected Behavior

Tests are working. 

### Current Behavior

What is the current behavior?

### Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

#### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. step 1
2. step 2
3. you get it...

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.13.1
* OS: macOS 12.6
* Kotlin version: 1.17.10
* JDK version: Java 19
* JUnit version: 5.9.2
* Type of test: unit test

#### Failure Logs

Please include any relevant log snippets or files here.

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
    io.mockk.MockKException: Missing mocked calls inside every { ... } block: make sure the object inside the block is a mock
        at app//io.mockk.impl.recording.states.StubbingState.checkMissingCalls(StubbingState.kt:14)
        at app//io.mockk.impl.recording.states.StubbingState.recordingDone(StubbingState.kt:8)
        at app//io.mockk.impl.recording.CommonCallRecorder.done(CommonCallRecorder.kt:47)
        at app//io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:64)
        at app//io.mockk.impl.eval.EveryBlockEvaluator.every(EveryBlockEvaluator.kt:30)
        at app//io.mockk.MockKDsl.internalEvery(API.kt:93)
        at app//io.mockk.MockKKt.every(MockK.kt:98)
[...]

Issue #864, opened 2022-07-26
Support Spying and Mocking HashMaps on JDK 16+


aSemy:
The tests in `HashMapMockTest.kt` fail on more recent (16+) JDK versions.

https://github.com/mockk/mockk/blob/02a6d4546d79da9d362b61376dae64ff656de7b1/mockk/jvm/src/test/kotlin/io/mockk/jvm/HashMapMockTest.kt#L36-L43

https://github.com/mockk/mockk/runs/7519107510?check_suite_focus=true#step:7:393

```none
HashMapMockTest > concurrentHashMap_shouldBeSpied_Successfully() FAILED
    java.lang.IllegalAccessException: class io.mockk.impl.InternalPlatform cannot access a member of class java.util.concurrent.ConcurrentHashMap (in module java.base) with modifiers "transient volatile"
        at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:[392])
        at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:674)
        at java.base/java.lang.reflect.Field.checkAccess(Field.java:1102)
        at java.base/java.lang.reflect.Field.get(Field.java:423)
        at io.mockk.impl.InternalPlatform.copyFields$copy(InternalPlatform.kt:114)
        at io.mockk.impl.InternalPlatform.copyFields(InternalPlatform.kt:121)
        at io.mockk.impl.instantiation.AbstractMockFactory.spyk(AbstractMockFactory.kt:107)
        at io.mockk.jvm.HashMapMockTest.concurrentHashMap_shouldBeSpied_Successfully(HashMapMockTest.kt:126)
```

This was discovered in https://github.com/mockk/mockk/pull/444#discussion_r523808327 and #829.

For now, so I can make progress on #863, I've disabled the tests on JDK16+. I'm making this issue to link in the source code.

omerates760:
@aSemy Hi, The problem continues. Is there an update?

Issue #834, opened 2022-06-16
@SpyK on java.time.Clock fails on Java 17


dc-up42:
### Expected Behavior

Using the following would lead to a spy of a `java.time.Clock` created.

```kotlin
    @SpyK
    private var clock = Clock.fixed(frozenInstant, ZoneId.of("UTC"))
```

Later it will be used to `@InjectMockKs`.

### Current Behavior

Tests fail to initialize. An `java.lang.IllegalAccessException` is thrown.

#### Steps to Reproduce

1. Write a test with a `@SpyK` over a `java.time.Clock`
2. Run the test

#### Context

* MockK version: 1.12.4
* Kotlin version: 1.7.0
* JDK version: openjdk 17.0.2 2022-01-18
* Type of test: unit test

#### Stack trace

<details>
<summary>
`java.lang.IllegalAccessException` thrown
</summary>

```
class io.mockk.impl.InternalPlatform$copyFields$1 cannot access a member of class java.time.Clock$SystemClock (in module java.base) with modifiers "private final"
java.lang.IllegalAccessException: class io.mockk.impl.InternalPlatform$copyFields$1 cannot access a member of class java.time.Clock$SystemClock (in module java.base) with modifiers "private final"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:392)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:674)
	at java.base/java.lang.reflect.Field.checkAccess(Field.java:1102)
	at java.base/java.lang.reflect.Field.get(Field.java:423)
	at io.mockk.impl.InternalPlatform$copyFields$1.invoke(InternalPlatform.kt:114)
	at io.mockk.impl.InternalPlatform.copyFields(InternalPlatform.kt:121)
	at io.mockk.impl.instantiation.AbstractMockFactory.spyk(AbstractMockFactory.kt:107)
	at io.mockk.impl.annotations.JvmMockInitializer.createSpyK(JvmMockInitializer.kt:127)
	at io.mockk.impl.annotations.JvmMockInitializer.assignSpyK(JvmMockInitializer.kt:117)
	at io.mockk.impl.annotations.JvmMockInitializer.initMock(JvmMockInitializer.kt:50)
	at io.mockk.impl.annotations.JvmMockInitializer.initAnnotatedMocks(JvmMockInitializer.kt:24)
[...]

Issue #681, opened 2021-08-11
JDK 16 & InaccessibleObjectException


knyttl:
### Expected Behavior

I upgraded to JDK16 & Gradle 7 and I have issues when running mockk:

```kotlin
// method for mocking timestamps
fun mockNow(string: String) {
    mockkStatic(Instant::class, ZonedDateTime::class, LocalDate::class, LocalDateTime::class)
    every { Instant.now() }.returns(instantOf(string))
    every { ZonedDateTime.now() }.returns(zonedDateTimeOf(string))
}

mockNow("2020-01-01T12:00:00Z")
```

Will fail with:

```
java.time.format.DateTimeParseException: Text '2020-01-01T12:00:00Z' could not be parsed: Unable to make private static java.time.LocalDate java.time.LocalDate.create(int,int,int) accessible: module java.base does not "opens java.time" to unnamed module @7254bd8e
        at java.base/java.time.format.DateTimeFormatter.createError(DateTimeFormatter.java:2023)
        at java.base/java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1958)
        at java.base/java.time.ZonedDateTime.parse(ZonedDateTime.java:600)
```

(see full stacktrace below)

### Current Behavior

We currently use JDK11 where this works correctly.

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.12.0
* OS: Any
* Kotlin version: 1.5.30-RC
* JDK version: 16
* JUnit version: 5.7.0
* Type of test: unit test

#### Full stacktrace
```
java.time.format.DateTimeParseException: Text '2020-01-01T12:00:00Z' could not be parsed: Unable to make private static java.time.LocalDate java.time.LocalDate.create(int,int,int) accessible: module java.base does not "opens java.time" to unnamed module @7254bd8e
        at java.base/java.time.format.DateTimeFormatter.createError(DateTimeFormatter.java:2023)
        at java.base/java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1958)
        at java.base/java.time.ZonedDateTime.parse(ZonedDateTime.java:600)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:78)
[...]

PR #965, opened 2022-11-07
Document InaccessibleObjectException/IllegalAccessException on JDK 16+ with solution

sergey-volkov-lm:
Some users face InaccessibleObjectException/IllegalAccessException working on JDK 16+.  
As far as I'm concerned, this is not considered as a bug in mockk; or fix is just not planned yet.  
So I added separate MD file describing the problem and link to it from main README.md (section "known issues")  

I'm not sure where to put this MD file, I supposed that `doc/md` directory is fine.

### Linked issues
* https://github.com/mockk/mockk/issues/681
* https://github.com/mockk/mockk/issues/929
* https://github.com/mockk/mockk/issues/864
* https://github.com/mockk/mockk/issues/834
* https://github.com/Ninja-Squad/springmockk/issues/65

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ From version 1.13.0 MockK supports Kotlin 1.4 and higher
  
 * PowerMock needs a workaround to run together with MockK [#79](https://github.com/mockk/mockk/issues/79#issuecomment-437646333). (not sure after workaround if it is generally usable or not, please somebody report it)
 * Inline functions cannot be mocked: see the discussion on [this issue](https://github.com/mockk/mockk/issues/27)
+* Spies, mockkStatic may not work on JDK 16+; `InaccessibleObjectException`/`IllegalAccessException`: [read more here](doc/md/jdk16-access-exceptions.md)
 
 Table of contents:
 
--- a/doc/md/jdk16-access-exceptions.md
+++ b/doc/md/jdk16-access-exceptions.md
@@ -1,0 +1,62 @@
+# JDK 16+ access exceptions
+
+### Problem symptoms 
+On JDK 16 and above you may encounter InaccessibleObjectException or IllegalAccessException in following known cases:
+1. _Some_ usages of `mockkStatic` on java standard library classes, e.g. 
+   ```kotlin
+   mockkStatic(Instant::class)
+   every { Instant.now() } returns Instant.parse("2022-09-15T09:31:26.919Z")
+   ```
+   ```
+   java.time.format.DateTimeParseException: Text '2022-09-15T09:31:26.919Z' could not be parsed:
+     Unable to make private static java.time.Instant java.time.Instant.create(long,int) accessible: 
+     module java.base does not "opens java.time" to unnamed module @4501b7af
+   ```
+
+2. Spying on java standard library classes, e.g.
+   ```kotlin
+   val socket = Socket()
+   val spy = spyk<Socket>(socket)
+   ```
+    Note: this includes creating `@SpyKBean` on spring data repository instances, because they are instances of `java.lang.reflect.Proxy` in fact.
+   ```
+   Error creating bean with name 'featureRepository': Post-processing of FactoryBean's singleton object failed;
+    nested exception is java.lang.IllegalAccessException: class io.mockk.impl.InternalPlatform 
+    cannot access a member of class java.lang.reflect.Proxy (in module java.base) with modifiers "protected"
+   ```
+
+### Problem cause
+JDK 16 enforced strong encapsulation of standard modules. In practice this means you (and your libraries) can no more get away with call like `privateMember.setAccessible(true)` to any of JDK classes.
+
+https://blogs.oracle.com/javamagazine/post/a-peek-into-java-17-continuing-the-drive-to-encapsulate-the-java-runtime-internals
+
+### Solution / workaround
+Add JVM argument `--add-opens java.base/java.time=ALL-UNNAMED` _for each package_ you need  
+Description: https://docs.oracle.com/en/java/javase/16/migrate/migrating-jdk-8-later-jdk-releases.html
+
+For gradle users:
+```groovy
+tasks.test {
+    jvmArgs("--add-opens", "java.base/java.time=ALL-UNNAMED")
+}
+```
+
+For maven users:
+```xml
+<plugin>
+    <groupId>org.apache.maven.plugins</groupId>
+    <artifactId>maven-surefire-plugin</artifactId>
+    <configuration>
+      <argLine>
+        --add-opens java.base/java.time=ALL-UNNAMED
+      </argLine>
+    </configuration>
+</plugin>
+```
+
+### Linked issues
+* https://github.com/mockk/mockk/issues/681
+* https://github.com/Ninja-Squad/springmockk/issues/65
+* https://github.com/mockk/mockk/issues/929
+* https://github.com/mockk/mockk/issues/864
+* https://github.com/mockk/mockk/issues/834


doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
...
+Description: https://docs.oracle.com/en/java/javase/16/migrate/migrating-jdk-8-later-jdk-releases.html
+
+For gradle users:
+```groovy
+tasks.test {
+    jvmArgs("--add-opens", "java.base/java.time=ALL-UNNAMED")
+}
+```
+
+For maven users:
```
aSemy, line 44:
```suggestion
For Maven users:
```

doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
...
+### Problem cause
+JDK 16 enforced strong encapsulation of standard modules. In practice this means you (and your libraries) can no more get away with call like `privateMember.setAccessible(true)` to any of JDK classes.
+
+https://blogs.oracle.com/javamagazine/post/a-peek-into-java-17-continuing-the-drive-to-encapsulate-the-java-runtime-internals
+
+### Solution / workaround
+Add JVM argument `--add-opens java.base/java.time=ALL-UNNAMED` _for each package_ you need  
+Description: https://docs.oracle.com/en/java/javase/16/migrate/migrating-jdk-8-later-jdk-releases.html
+
+For gradle users:
```
aSemy, line 37:
```suggestion
For Gradle users:
```

doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
+# JDK 16+ access exceptions
+
+### Problem symptoms 
+On JDK 16 and above you may encounter InaccessibleObjectException or IllegalAccessException in following known cases:
+1. _Some_ usages of `mockkStatic` on java standard library classes, e.g.
```
aSemy, line 5:
```suggestion
1. _Some_ usages of `mockkStatic` on Java standard library classes, e.g. 
```

doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
...
+   mockkStatic(Instant::class)
+   every { Instant.now() } returns Instant.parse("2022-09-15T09:31:26.919Z")
+   ```
+   ```
+   java.time.format.DateTimeParseException: Text '2022-09-15T09:31:26.919Z' could not be parsed:
+     Unable to make private static java.time.Instant java.time.Instant.create(long,int) accessible: 
+     module java.base does not "opens java.time" to unnamed module @4501b7af
+   ```
+
+2. Spying on java standard library classes, e.g.
```
aSemy, line 16:
```suggestion
2. Spying on Java standard library classes, e.g.
```

doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
+# JDK 16+ access exceptions
+
+### Problem symptoms 
+On JDK 16 and above you may encounter InaccessibleObjectException or IllegalAccessException in following known cases:
```
aSemy, line 4:
```suggestion
On JDK 16 and above you may encounter `InaccessibleObjectException` or `IllegalAccessException` in following known cases:
```

README.md
```diff
@@ -33,6 +33,7 @@ From version 1.13.0 MockK supports Kotlin 1.4 and higher
  
 * PowerMock needs a workaround to run together with MockK [#79](https://github.com/mockk/mockk/issues/79#issuecomment-437646333). (not sure after workaround if it is generally usable or not, please somebody report it)
 * Inline functions cannot be mocked: see the discussion on [this issue](https://github.com/mockk/mockk/issues/27)
+* Spies, mockkStatic may not work on JDK 16+; `InaccessibleObjectException`/`IllegalAccessException`: [read more here](doc/md/jdk16-access-exceptions.md)
```
aSemy, line 36:
```suggestion
* Spies, `mockkStatic` may not work on JDK 16+; `InaccessibleObjectException`/`IllegalAccessException`: [read more here](doc/md/jdk16-access-exceptions.md)
```

aSemy:
Ahhhhh is that why it's been so tricky to create a unit test in the project for complete JDK 17 support? Because the test classes are allowed to create instances of classes within the same package?

doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
...
+   ```
+    Note: this includes creating `@SpyKBean` on spring data repository instances, because they are instances of `java.lang.reflect.Proxy` in fact.
+   ```
+   Error creating bean with name 'featureRepository': Post-processing of FactoryBean's singleton object failed;
+    nested exception is java.lang.IllegalAccessException: class io.mockk.impl.InternalPlatform 
+    cannot access a member of class java.lang.reflect.Proxy (in module java.base) with modifiers "protected"
+   ```
+
+### Problem cause
+JDK 16 enforced strong encapsulation of standard modules. In practice this means you (and your libraries) can no more get away with call like `privateMember.setAccessible(true)` to any of JDK classes.
```
aSemy, line 29:
```suggestion
JDK 16 enforces strong encapsulation of standard modules. In practice this means libraries are forbidden from overriding accessibility modifiers in JDK classes (for example,  `privateMember.setAccessible(true)`).
```

doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
...
+
+https://blogs.oracle.com/javamagazine/post/a-peek-into-java-17-continuing-the-drive-to-encapsulate-the-java-runtime-internals
+
+### Solution / workaround
+Add JVM argument `--add-opens java.base/java.time=ALL-UNNAMED` _for each package_ you need
```
aSemy, line 34:
```suggestion
Add JVM argument `--add-opens java.base/java.time=ALL-UNNAMED` _for each package_ you need to mock.
```

aSemy (approved):
Nice work :)

I have some minor formatting and grammar tweaks.

I'm approving, but I'm not certain how the website works - but that can be fixed later if the file isn't accessible.

doc/md/jdk16-access-exceptions.md
sergey-volkov-lm: > In practice this means libraries are forbidden

AFAIK not only libraries. Do you want to make accent on mockk use-case?

doc/md/jdk16-access-exceptions.md
sergey-volkov-lm: Agree.
And to be more general here, we also can replace `java.time` with `package.name`. And there are specific examples later in the file

doc/md/jdk16-access-exceptions.md
aSemy: Good point, I've edited it. 

I think the project name should be styled as: MockK

doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
...
+### Solution / workaround
+Add JVM argument `--add-opens java.base/java.time=ALL-UNNAMED` _for each package_ you need  
+Description: https://docs.oracle.com/en/java/javase/16/migrate/migrating-jdk-8-later-jdk-releases.html
+
+For Gradle users:
+```groovy
+tasks.test {
+    jvmArgs("--add-opens", "java.base/java.time=ALL-UNNAMED")
+}
+```
```
sergey-volkov-lm, line 42: I am not very familiar with gradle, and copied this from one of the issues. May need help from experienced gradle user to verify that this snippet looks ok

aSemy: Yeah it looks okay. It will only change the 'test' class though. In .kts, this is the best way to change all test tasks:

```kts
// build.gradle.kts

tasks.withType<Test>().configureEach {
  jvmArgs(...)
}
```

doc/md/jdk16-access-exceptions.md
aSemy: Looking at the [docs here](https://docs.oracle.com/en/java/javase/16/migrate/migrating-jdk-8-later-jdk-releases.html#GUID-2F61F3A9-0979-46A4-8B49-325BA0EE8B66), would the flag `--illegal-access=warn` work better?

sergey-volkov-lm: They say `--illegal-access` was removed in JDK 17

[Article](https://blogs.oracle.com/javamagazine/post/a-peek-into-java-17-continuing-the-drive-to-encapsulate-the-java-runtime-internals), section "What’s changing in Java 17?"

Because of this I did not mention this flag. I think that JDK 17 is going to be much more popular, and current users of JDK 16 should not rely on `--illegal-access` option, in order not to face access problems again in future.

doc/md/jdk16-access-exceptions.md
```diff
@@ -0,0 +1,62 @@
+# JDK 16+ access exceptions
+
+### Problem symptoms 
+On JDK 16 and above you may encounter `InaccessibleObjectException` or `IllegalAccessException` in following known cases:
```
Raibaz, line 4: "in the following known cases"

Raibaz (approved):
This is brilliant, thanks a lot!

Added just a minor grammar nit and I agree with @aSemy's comments and clarifications, but then we can merge it.

I'm not 101% familiar with the website generation so I don't know if linking to a .md file will actually render it or just show the markdown, but I can fix it once we've merged this PR.

Thanks again!

doc/md/jdk16-access-exceptions.md
aSemy: Ah clear, thanks for explaining.

aSemy:
Is it possible for MockK to detect, during runtime, if the correct flags are set? Then MockK could print a warning, directing users to these docs.

sergey-volkov-lm:
Sorry for delay. I have fixed all what was clear to me. And added additional details because of misunderstanding here https://github.com/mockk/mockk/issues/981

Raibaz:
Thanks!

PR #979, opened 2022-11-22
Update README.md to include note about 2-year-old 'known issue' for spy and suspend function

iainism:
Added note (to spy and coroutines sections) about known issue with suspending functions and spy (issue now nearly two years old)


Raibaz:
Thanks!

Issue #978, opened 2022-11-18
MockKException: Failed matching mocking signature with mocked sealed type on 1.13.2


feerbach:
#### Steps to Reproduce

1. Mock sealed interface or class
2. Setup the created mock call using the `every { ... }` block (I used the `any()` matcher for argument)

#### Context

* MockK version: 1.13.2
* Kotlin version: 1.7.21
* JDK version: 11
* JUnit version: 5
* Type of test: unit test

#### Stack trace

```
io.mockk.MockKException: Failed matching mocking signature for

left matchers: [any()]

	at io.mockk.impl.recording.SignatureMatcherDetector.detect(SignatureMatcherDetector.kt:99)
	at io.mockk.impl.recording.states.RecordingState.signMatchers(RecordingState.kt:39)
	at io.mockk.impl.recording.states.RecordingState.round(RecordingState.kt:31)
	at io.mockk.impl.recording.CommonCallRecorder.round(CommonCallRecorder.kt:50)
	at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:63)
	at io.mockk.impl.eval.EveryBlockEvaluator.every(EveryBlockEvaluator.kt:30)
	at io.mockk.MockKDsl.internalEvery(API.kt:93)
	at io.mockk.MockKKt.every(MockK.kt:98)
...

```

#### Minimal reproducible code (the gist of this issue)

```kotlin
import io.mockk.*
import org.junit.jupiter.api.Test
import kotlin.test.assertNotNull

sealed interface Sealed {
    abstract fun doSomething(arg: Any?)
}

class A : Sealed {
    override fun doSomething(arg: Any?) {}
}

class SealedClassTest {

    @Test
    fun test() {
        val a = mockk<Sealed>()

        every { a.doSomething(any()) } just Runs

        assertNotNull(a)
    }
}
```

PR #980, opened 2022-11-25
Prevent "MockKException: Failed matching mocking signature for" on sealed classes

diogopmonteiro:
Hello, I was trying to upgrade a Kotlin project I'm working on to kotlin 1.7.21 (jdk 17) and was having the same problem as described in https://github.com/mockk/mockk/issues/978.

This PR is an attempt to fix the issue. I'll try to explain what I found out, but please keep in mind that I know very little what I'm talking about 😂 . 

The exception reported in the issue is thrown [here](https://github.com/mockk/mockk/blob/master/modules/mockk/src/commonMain/kotlin/io/mockk/impl/recording/SignatureMatcherDetector.kt#L99), because a single `CallRound` arrives in `detect(callRounds: List<CallRound>)` and: 
- has a single `SignedMatcher` (indicating that `any()` was called), 
- but has no `SignedCall`s to match with the existing `SignedMatcher`. 

From what I could understand `SignedCall`s are registered [here](https://github.com/mockk/mockk/blob/master/modules/mockk/src/commonMain/kotlin/io/mockk/impl/recording/states/RecordingState.kt#L64-L90), when the method call is intercepted in the proxy call (i.e. when a mocked method is called inside a `every { }` block). For sealed classes proxies, this "interception" is not happening.

I think this "interception" is configured by calling [inline](https://github.com/mockk/mockk/blob/master/modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ProxyMaker.kt#L30) on the mock class. It happens that sealed class proxies are actually an instance of a sealed subclass (see [here](https://github.com/mockk/mockk/blob/master/modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ProxyMaker.kt#L110)). So, `mockk` instantiates a sealed subclass, but calls `inline` on the sealed class (its parent). 

The change in this PR is to call `inline` on the same sealed subclass that `mockk` instantiates. This seems to correctly register the interceptors, which in turn correctly store the `SignedCall` in `RecordingState`, which prevents the exception reported in the issue.

I tested this change on a project with JDK 17 and Kotlin 1.7.21, and seems to be working. Didn't test in other JDKs/Kotlin versions.

Does any of this make sense? 😂

### DIFF
--- a/modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ProxyMaker.kt
+++ b/modules/mockk-agent/src/jvmMain/kotlin/io/mockk/proxy/jvm/ProxyMaker.kt
@@ -80,15 +80,20 @@ internal class ProxyMaker(
     private fun <T : Any> inline(
         clazz: Class<T>
     ): () -> Unit {
-        val superclasses = getAllSuperclasses(clazz)
+        val resolvedClass = if (clazz.kotlin.isSealed) {
+            clazz.kotlin.sealedSubclasses.firstOrNull()?.java
+                ?: error("Unable to resolve subclass for sealed class $clazz, available subclasses: ${clazz.kotlin.sealedSubclasses}")
+        } else clazz
+
+        val superclasses = getAllSuperclasses(resolvedClass)
 
         return if (inliner != null) {
             val transformRequest = TransformationRequest(superclasses, SIMPLE)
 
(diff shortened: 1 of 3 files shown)


diogopmonteiro:
That is fair. Thanks for the patch. I'll close this PR.

Issue #989, opened 2022-12-01
SLF4j 2.0.x Support with Log4j2


desilvai-target:
### Summary
From #243, mockk checks if SLF4j is available for use as a logger by checking the classpath for `org.slf4j.impl.StaticLoggerBinder`.  However, as of SLF4j 2.x, this file doesn't exist in the SLF4j implementations (we checked logback and the `log4j-slf4j2-impl` implementations).  This can lead to the incorrect logger being used, or, if the wrong connector is on the classpath, an exception being thrown.

### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [X] I am running the latest version
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed

### Expected Behavior
SLF4j 2.x loggers are used by mockk if present in classpath.

### Current Behavior
mockk cannot locate the SLF4j 2.x logger.  This disables logging within our tests.  

When we include packages with transitive dependencies on an older SLF4j implementation, this causes a `NoSuchMethodError` during mock initialization. 

### Solution
The line at issue is in `io.mockk.impl.log.JvmLogging::slf4jOrJulLogging()`.  Rather than checking for files in the classpath, we can attempt to create an SLF4j logger for any class and if it fails, fall back to the JULLogger.  This will allow mockk to work with both SLF4j 1.7.x and 2.x and be resistant to future SLF4j changes.

Raibaz:
Sorry for the delay, v1.13.4 is out now.

Slava-Miligoulo-Target:
Awesome, wrapped updating things on our and it's working great.  Thanks!

desilvai-target:
As Slava said, it is working great on our end.  Thank you both!  Really appreciate both of your work on this!

jhackel-hypo:
This seems to break tests in Spring Boot 2.x applications. Since we updated to version 1.13.4 of mockk, nothing is logged anymore for out tests. Instead we get this error:

```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
SLF4J: Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
[...]

PR #990, opened 2022-12-01
fix: [#989] Slf4J 2.0+ Support

Slava-Miligoulo-Target:
Version 2.0+ Slf4j , `org.slf4j.impl.StaticLoggerBinder` has been removed.  Additional info at #989.

Adjusted slf4jOrJulLogging to attempt to create Slf4j instance instead of checking class path and, if unable, default to JULLogger.

logger-tests sub module added to test without slf4j-impl in classpath.


Raibaz:

Will merge as soon as the build passes.

PR #1003, opened 2022-12-18
Create a Lint detector to warn about undesired usages of mockk

SimonMarquis:
And publish it to be consumed by clients of the `mockk-android` dependency.

### DIFF
--- a/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
+++ b/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
@@ -18,6 +18,7 @@ object Deps {
         const val junitJupiter = "5.8.2"
         const val junitVintage = "5.8.2"
         const val junit4 = "4.13.2"
+        const val lint = "30.3.1"
 
         const val byteBuddy = "1.12.20"
         const val objenesis = "3.3"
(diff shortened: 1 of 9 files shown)


SimonMarquis:
👋 Hi, I'd like to get some feedback on this kind of Lint detector, especially regarding "undesired" mockk usages.
It currently detects and warns against mocks of data classes, enums, primitives and interfaces. 
Also, I think the `Issue`'s `briefDescription` and `explanation` wording could be improved a bit.

Raibaz:

However, I have a few issues:
- Is there a way to make this not specific for Android? Mocking a data class is something we should detect and advise against on all platforms, not just on Android
- Is this configurable? I.e., do users have an easy way to enable and disable specific detections? Ideally, I'd like this to be opt-in, so that users have to explicitly say `please warn me when someone is mocking a data class and tell them they shouldn't`, rather then introducing a different behavior for all existing code
- (Related to the previous point) I'm not 101% convinced we should advise against mocking interfaces

SimonMarquis:
> Is there a way to make this not specific for Android?

While Lint is not really an Android specific thing (it has been built initially outside Android), I currently don't know any way to use them outside Android projects. Also, the publishing step requires the AGP plugin to do all the necessary plumbing (`lintPublish` configuration) and the output `lint.jar` is actually shipped into and AAR (Android archive format) and consumer automatically picks it up.
And I don't know any other target-agnostic tool that would be already baked in.
The alternatives are external linters like Detekt or ktlint, but this would force consumer to include them in their project. Though it might be good to have options. (ktlint custom rule seems doable but I've never used it).

> Is this configurable

Yes it is, and as you mentioned it, I'll make them opt-in.
The opt-in step will be something like:

```kotlin
android {
  lintOptions {
    // opt-in single rule
    enable += "MockkDataClass"
    // opt-in multiple rules
    enable += setOf("MockkDataClass", "MockkEnum", "MockkInterface", "MockkPrimitive")
  }
}
```

PR #1004, opened 2022-12-18
Customize jobs name to make the matrix values more explicit

SimonMarquis:
For instance, instead of the default `tests (11, 1.5.31, true)`, it will render as `[java=11, kotlin=1.5.31, IR=true]`.


SimonMarquis:
https://github.com/mockk/mockk/actions/runs/3726213422 to view the result.

Raibaz:
Nice one, thanks!

PR #1005, opened 2022-12-18
Add `fun <T> MockKMatcherScope.any(KClass<T>): T`

ghackett:
### Premise

In transitioning from mockito to mockk, the one feature I've been missing is Mockito's `defaultAnswer` concept. 

It seems like we can actually replicate this concept in mockk using reflection, however we're blocked w/o a way to define an `any()` matcher given a KClass or KType instead of a reified type.

### The Change

This PR currently adds only the needed `fun any(KClass)` and `fun match(Matcher, KClass)` functions to the api to open up the possibility of mockking via reflection, and avoids trying to define new dsl for declaring default answers. 

We also demonstrate two examples of reflective mockking. In both examples we mock a TestBuilder and tell it to always return itself. This is just to demonstrate the possibilities we'd be opening up with this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1209,6 +1209,24 @@ inline fun <reified T : List<E>, E : Any> MockKMatcherScope.matchListWithoutOrde
 ): T = match(ListWithoutOrderMatcher(listOf(*items), refEq))
 ```
 
+### Reflection matchers
+
+Example using reflection to mock all methods on a builder-style object
+
+```kotlin
+val builderFunctions = MyBuilder::class.memberFunctions.filter { it.returnType.classifier == MyBuilder::class }
+val builderMock = mockk<MyBuilder> {
+  builderFunctions.forEach { func ->
+    every {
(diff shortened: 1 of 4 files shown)


nicchristeller:
I've been trying to do very similar, and came to the same conclusion. 

`fun <T> MockKMatcherScope.any(KClass<T>): T` would be super useful.

ghackett:
👋 For the mockk maintainers, is this kind of change something that could be considered? If so, what does this PR need to become acceptable/mergable? I'm happy to put in the work, just need some guidance on what's missing.

ghackett:
In the meantime, I went ahead and shipped a library that enables this functionality today and **does** attempt to define a DSL for reflective mocking. Check it out at https://episode6.github.io/reflective-mockk/

Kevinrob (approved):

ghackett:
Awesome! Rebased and updated with a readme entry. Thanks @Raibaz!

Issue #1016, opened 2023-01-02
Matcher invoked multiple times unexpectedly


camhashemi:
### Expected Behavior

`match` and `withArg` should be invoked exactly once per test invocation (right?).

### Current Behavior

After upgrading to mockk 1.13.3, lamba blocks for `match` and `withArg` are invoked multiple times.

This is generally fine (though perhaps slow and unnecessary) for stateless tests, but it fails for tests that match against stateful objects like a `FileInputStream`.

### Failure Information (for bugs)

This doesn't happen on mockk 1.13.1.

#### Steps to Reproduce

See code sample

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 13.1.3
* OS: Mac
* Kotlin version: 1.7.22
* JDK version: 17
* JUnit version: 5
* Type of test: unit test

#### Failure Logs

```
getting ready to verify...
verifying...
verified!
```

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
match called twice!
java.lang.Exception: match called twice!
	at com.traderepublic.card.settlement.MockkTest$reproduce verify called multiple times$1$1.invoke(MockkTest.kt:26)
	at com.traderepublic.card.settlement.MockkTest$reproduce verify called multiple times$1$1.invoke(MockkTest.kt:25)
	at io.mockk.FunctionMatcher.match(Matchers.kt:63)
	at io.mockk.InvocationMatcher.match(API.kt:3878)
	at io.mockk.impl.verify.UnorderedCallVerifier.matchCall(UnorderedCallVerifier.kt:103)
	at io.mockk.impl.verify.UnorderedCallVerifier.verify(UnorderedCallVerifier.kt:38)
	at io.mockk.impl.recording.states.VerifyingState$recordingDone$outcome$1.invoke(VerifyingState.kt:31)
	at io.mockk.impl.recording.states.VerifyingState$recordingDone$outcome$1.invoke(VerifyingState.kt:30)
	at io.mockk.impl.recording.CommonCallRecorder.safeExec(CommonCallRecorder.kt:72)
	at io.mockk.impl.recording.states.VerifyingState.recordingDone(VerifyingState.kt:30)
	at io.mockk.impl.recording.CommonCallRecorder.done(CommonCallRecorder.kt:47)
	at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:64)
[...]

PR #1017, opened 2023-01-03
Don't call match an extra time during verification

camhashemi:
Fixes https://github.com/mockk/mockk/issues/1016


Raibaz:
Thanks a lot for looking into this! :)

Issue #1013, opened 2022-12-30
Spyk'ed object can't be  unmocked which leads to OutOfMemoryError


bakomchik:
### Expected Behavior
unmock(spyk()) -- works
No `java.lang.OutOfMemoryError: Java heap space` error occurred.

### Current Behavior

`java.lang.OutOfMemoryError: Java heap space` occurs

`unmock(spyk())` -- does not unmock spyk'ed object

#### Context

We have a lot of SpringBoot tests . 
We annotated `@Lazy` bean with `@SpykBean` in our TestCaseBase class
And sometimes new spring context created and bean get `spyked` again and it leads to OutOfMemoryError
because spiked object can't be unmocked/unspyked

* MockK version:     ` implementation("io.mockk:mockk:1.13.3")`
* OS:  `Darwin 21.6.0 Darwin Kernel Version 21.6.0: Thu Sep 29 20:13:56 PDT 2022; root:xnu-8020.240.7~1/RELEASE_ARM64_T6000 arm64`

* Kotlin version:  `1.7.21`
* JDK version: `openjdk 11.0.16.1 2022-08-12
OpenJDK Runtime Environment Homebrew (build 11.0.16.1+0)
OpenJDK 64-Bit Server VM Homebrew (build 11.0.16.1+0, mixed mode)
`
* Type of test: unit test

#### Minimal reproducible code (the gist of this issue)

```kotlin
// -----------------------[ GRADLE DEFINITIONS ] -----------------------
dependencies {
    testImplementation(kotlin("test"))
    implementation("io.mockk:mockk:1.13.3")
}

tasks.test {
    useJUnitPlatform()
}

tasks.withType<KotlinCompile> {
    kotlinOptions.jvmTarget = "1.8"
}

application {
    mainClass.set("MainKt")
}
// -----------------------[ YOUR CODE STARTS HERE ] -----------------------


fun main(args: Array<String>) {
    (0..1000).forEach {
        val spyk = spyk(LazySpringBeanWhichHoldsReferenceForBeanFactory(ByteArray(10 * 1024 * 1024)))
        every { spyk.doSmth() } returns "Wow"
        spyk.doSmth();
        clearMocks(spyk)
        unmockkObject(spyk)
    }
}

class LazySpringBeanWhichHoldsReferenceForBeanFactory(val beanFactory: ByteArray){
    fun doSmth():String{
            return "Smth"
    }
}
```

bakomchik:
Path to gc roots
[...]

PR #1018, opened 2023-01-04
#1013 uncouple proxy objects from cancellable to make proxy available…

bakomchik:
uncouple proxy objects from cancellable to make proxy availabl to GC


Raibaz:
Thanks for looking into this!

Issue #223, opened 2019-01-11
Bug: unable to capture arguments passed to generic functions


TAGC:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

I'm trying to create a mock containing a generic function, and I'd like to capture the arguments it receives. For example, like this:

```kotlin
val receivedValues = mutableListOf<T>()
val receiver = mockk<InputPort<T>> {
    every { id } returns "Receiver"
    coEvery { accept(capture(receivedValues)) } coAnswers { yield() }
}
```

I'd expect that if I create (for example) a mock instance of `InputPort<Int>`, then I should be able to call `receiver.accept(3)` and have `3` captured without any error.

### Current Behavior

Trying to call the function `receiver.accept` with any argument results in `io.mockk.MockKException: no answer found for: InputPort(#1).accept(10, continuation {})`.

It doesn't matter whether the function is suspending or not.

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.9
* OS: Windows 10 (64-bit)
* Kotlin version: 1.3.11
* JDK version: 1.8.0_191
* JUnit version: 5
* Type of test: unit test

#### Stack trace

```
Mockk failure invoking 'bar(3)': no answer found for: Foo(#1).bar(3)
Exception in thread "main" io.mockk.MockKException: no answer found for: Foo(#1).bar(3)
	at io.mockk.impl.stub.MockKStub.defaultAnswer(MockKStub.kt:99)
	at io.mockk.impl.stub.MockKStub.answer(MockKStub.kt:51)
	at io.mockk.impl.recording.states.AnsweringState.call(AnsweringState.kt:16)
	at io.mockk.impl.recording.states.AnsweringStillAcceptingAnswersState.call(AnsweringStillAcceptingAnswersState.kt:16)
	at io.mockk.impl.recording.CommonCallRecorder.call(CommonCallRecorder.kt:54)
	at io.mockk.impl.stub.MockKStub.handleInvocation(MockKStub.kt:272)
	at io.mockk.impl.instantiation.JvmMockFactoryHelper$mockHandler$1.invocation(JvmMockFactoryHelper.kt:25)
	at io.mockk.proxy.jvm.advice.Interceptor.call(Interceptor.kt:20)
[...]

PR #1029, opened 2023-01-20
Edit CapturingGenericArgumentsTest

bossm0n5t3r:
## Describe your changes

- Delete test Ignore annotation
- Delete test comment
- Add private modifier to some methods
- Use single name import instead of wildcard import

### DIFF
--- a/modules/mockk/src/commonTest/kotlin/io/mockk/it/CapturingGenericArgumentsTest.kt
+++ b/modules/mockk/src/commonTest/kotlin/io/mockk/it/CapturingGenericArgumentsTest.kt
@@ -1,15 +1,13 @@
 package io.mockk.it
 
-import io.mockk.*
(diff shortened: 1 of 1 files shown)


modules/mockk/src/commonTest/kotlin/io/mockk/it/CapturingGenericArgumentsTest.kt
```diff
@@ -1,15 +1,13 @@
...
+import io.mockk.mockk
+import io.mockk.slot
 import kotlin.test.Test
 
-@Ignore
 class CapturingGenericArgumentsTest {
-    /**
-     * Unable to capture arguments passed to generic functions.
-     * Verifies issue #223.
-     */
```
bossm0n5t3r, line 12: issue #223 is here

PR #1032, opened 2023-01-27
Return the configured answers when no answer found

maeserichar:
This PR modifies the way the `MockKException` is constructed when a mock is called with a set of parameters that wasn't configured, so it also adds to the exception message all the answers configured for that mock.

The aim of this change is to help understand why is the mock's call really failing, giving some extra information to the developer.

### DIFF
--- a/modules/mockk/src/commonMain/kotlin/io/mockk/impl/stub/MockKStub.kt
+++ b/modules/mockk/src/commonMain/kotlin/io/mockk/impl/stub/MockKStub.kt
@@ -88,8 +88,8 @@ open class MockKStub(
                 }
             } else {
                 val configuredAnswers = answers.map { it.matcher.toString() }.joinToString(separator = "\n") { it }
-                throw MockKException("no answer found for: ${gatewayAccess.safeToString.exec { invocation.toString() }}" +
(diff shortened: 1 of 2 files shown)


modules/mockk/src/commonMain/kotlin/io/mockk/impl/stub/MockKStub.kt
```diff
@@ -87,7 +87,9 @@ open class MockKStub(
                     childMockK(invocation.allEqMatcher(), invocation.method.returnType)
                 }
             } else {
-                throw MockKException("no answer found for: ${gatewayAccess.safeToString.exec { invocation.toString() }}")
+                val configuredAnswers = answers.map { it.matcher.toString() }.joinToString(separator = "\n") { it }
+                throw MockKException("no answer found for: ${gatewayAccess.safeToString.exec { invocation.toString() }}" +
```
Raibaz, line 91: I would remove the first semicolon here and replace "between" with "among"

Raibaz (changes requested):

Just a few nits on the format of the log message but otherwise LGTM

Raibaz (approved):

PR #1034, opened 2023-01-29
replace "no answer found for" by "no mock found for"

ELBEQQAL94:
it's a misleading error for beginners.


Raibaz:
I don't think this is correct; you can have a mock defined for an object but no answer for a method, and this exception is being thrown when no answer is defined, so the existing message is right.

Issue #1037, opened 2023-02-02
MockK unbox value class return value for mocked method


jijichen:
> Just be aware about usual MockK support pattern. 
> Tickets are checked from time to time, replied, discussed, labeled, e.t.c.
> But real fixes are applied in a month-two month period in a bunch.
> If you think this is unacceptable, go on, join the project, change the world.

## Please remove sections wisely

Below information is actually needed to make all the process of fixing faster.
Choose main points. Don't blindly follow this as a set of rules. 
Don't waste much time. Usually, the main thing is to have a good reproducible minimal code.

### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

when moccking following class's foo function
```
class Api {
    fun foo() : Any
}
```

with `every { mockApi.foo() } returns Result.success(2)`, I would think the return of mocked method would be `Result.sucess(2)`

### Current Behavior

The mockked return is unboxed to be 2. 

### Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

#### Steps to Reproduce

1. Mock a method that returns 'Any', to return `Result.success(100)`
2. assert on the value of mocked return.
3.  mocked return is unboxed.

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.13.4
* OS: macos
* Kotlin version: 1.8.0
* JDK version: 18.0.2
* JUnit version: 5.6.3
* Type of test: unit test

#### Failure Logs

Please include any relevant log snippets or files here.

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
expected: Success(2)
 but was: 2
org.opentest4j.AssertionFailedError: 
expected: Success(2)
 but was: 2
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at MockkTest.mockked and failed(MockkTest.kt:25)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #1038, opened 2023-02-02
Issue#1037 : Don't unbox Result inline class for interceptor

jijichen:
In issue #1037 , there appears to be an issue with unbox Result class, causing the return value type of Result class, being unboxed into underlying value type.

This was a bit confusing and break existing old tests on my side.


jijichen:
Already fixed by #1204

PR #1046, opened 2023-02-10
fix: [#989] Slf4J and logback downgrade

Slava-Miligoulo-Target:
Lowered baseline slf4j and logback version for backward compatability.


desilvai-target (approved):
This will fix backward compatibility issues mentioned in comments for #989

Raibaz:
I don't think we should depend on an unsupported version of SLF4J.

I'r dather keep depending on the latest version and have people exclude our own dependency they really want to depend on older ones.

desilvai-target:
> I don't think we should depend on an unsupported version of SLF4J.
> 
> I'd rather keep depending on the latest version and have people exclude our own dependency they really want to depend on older ones.

That would be my preference too.  Just wasn't sure if the project objectives were around minimal changes for the user or staying somewhat current.

Slava-Miligoulo-Target:
Works for us, going to close out this PR.

Issue #997, opened 2022-12-10
Can't instantiate proxy for class kotlin.Function2 - similar class definition


rgonciarz:
JDK 19
Quarkus 2.14.3.Final
Mockk 1.13.3
JUnit 5.9.1

I'm trying to replace an existing instance with mock:

`typealias BroadcastSwitchActiveComputer = suspend (Computer) -> Either<BroadcastSwitchActiveComputerError, Unit>`

`kotlin.jvm.functions.Function2<? super Computer, ? super kotlin.coroutines.Continuation<? super arrow.core.Either<? extends BroadcastSwitchActiveComputerError, ? extends kotlin.Unit>>, ?>`

I have several beans with similar Function2 class definition. However they have different Either's generic type.
 

```
val typeLiteral = object : TypeLiteral<BroadcastSwitchActiveComputer>() {}
private val mockInstance: BroadcastSwitchActiveComputer = mockk()
val beanInstance = CDI.current().select(typeLiteral, Named("BroadcastSwitchActiveComputer"))
QuarkusMock.installMockForInstance(mockInstance, beanInstance);

```

A full stack trace:

```
Failed to create test instance
org.junit.jupiter.api.extension.TestInstantiationException: Failed to create test instance
	at io.quarkus.test.junit.QuarkusTestExtension.initTestState(QuarkusTestExtension.java:773)
	at io.quarkus.test.junit.QuarkusTestExtension.interceptTestClassConstructor(QuarkusTestExtension.java:737)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.api.extension.InvocationInterceptor.interceptTestClassConstructor(InvocationInterceptor.java:73)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
[...]

PR #1052, opened 2023-02-15
Issue #997 - Simulate garbage collected proxy

ygaller:
Fix for issue #997 - WeakReference fails when GC is called between `proxy` and `get`, resulting in flaky tests. This PR simulates & solves the issue.


gmazzo:
Any chance to get this merged soon? We just confirmed running our CI with this patched versions fixes our 20% instability issue on JDK17:
`io.mockk.proxy.MockKAgentException: Value for this result is not assigned`

PR #1056, opened 2023-02-19
update Kotlin and Java Toolchain properties, and setting them in CI

### LINKED ISSUE #1048: Kotlin version configured from GitHub Actions workflow is not correctly applied
CI uses the following matrix to test Kotlin versions:

https://github.com/mockk/mockk/blob/6c7a06cedafc6b00b127a46ea304f404dcb3c180/.github/workflows/gradle.yml#L20

and propagate them as Gradle project properties:

https://github.com/mockk/mockk/blob/6c7a06cedafc6b00b127a46ea304f404dcb3c180/.github/workflows/gradle.yml#L41

MockK uses this property here:

https://github.com/mockk/mockk/blob/6c7a06cedafc6b00b127a46ea304f404dcb3c180/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt#L6

but this method is unused in this project.
And therefore I suspect all jobs are using the same Kotlin version.

---

Also, I tried to run the same command as in the workflow but with an invalid Kotlin version, and it still ran the tests sucessfully:

```bash
gradlew check --stacktrace -Pkotlin.version=0.0.0 -Pkotlin.ir.enabled=true -PjavaToolchainTestVersion=17
```

aSemy:
Fix #1048 

* prefix the Java and Kotlin versions with `io_mockk`, which can help in some situations (e.g. if an individual wants to set them specifically on their machine they can add them in `$GRADLE_USER_HOME/gradle.properties` and not risk clashing with other properties)
* standardise the naming for the Kotlin and Java versions (I chose snake_case, but camelCase would work as well - up to you)
* remove Java version/target/toolchain from `buildSrc/build.gradle.kts` - it should be inherited from the root project

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -34,11 +34,12 @@ jobs:
         uses: gradle/gradle-build-action@v2.4.2
 
       - name: Run tests with Gradle
+        env:
+          ORG_GRADLE_PROJECT_io_mockk_kotlin_version: ${{ matrix.kotlin-version }}
+          ORG_GRADLE_PROJECT_io_mockk_java_toolchain_test_version: ${{ matrix.java-version }}
         run: >
           ./gradlew check
           --stacktrace
-          -Pkotlin.version=${{ matrix.kotlin-version }}
-          -PjavaToolchainTestVersion=${{ matrix.java-version }}
 
   android-instrumented-tests:
     runs-on: macos-latest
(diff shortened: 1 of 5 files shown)


Raibaz:
Is this still needed? @aSemy if so, can you please resolve the conflicts?

aSemy:
hey @Raibaz, it's been a while since I looked at this, so I think this PR is now just a bit of tidying.

* prefix the Java and Kotlin versions with `io_mockk`, so individuals can set them in `$GRADLE_USER_HOME/gradle.properties` and not risk clashing with other properties
* standardise the naming for the Kotlin and Java versions (I chose snake_case, but camelCase would work as well - up to you)

PR #1061, opened 2023-03-06
Added the "Top Level functions" section

polarene:
Explained how to mock top-level functions. Added a code example using the mockkStatic(KFunction<*>) overload.

### DIFF
--- a/README.md
+++ b/README.md
@@ -877,6 +877,24 @@ runTest {
 
 Note: there is a known issue if using a spy with a suspending function: https://github.com/mockk/mockk/issues/554
 
+### Top Level functions
+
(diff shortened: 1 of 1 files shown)


Raibaz:
Thanks!

mcclurej06:
This change states that:
>Mocking a function doesn't affect other functions declared in the same file.

Which seems to contradict the documentation here:
https://github.com/mockk/mockk/blob/0fa0b267aa1f4a5b1b37b5b3bec45c14f78d1adb/modules/mockk/src/jvmMain/kotlin/io/mockk/MockK.kt#L23-L27

polarene:
You're right, it isn't the best wording choice... it doesn't express exactly what I meant. To start with I wasn't aware of this behavior, what I meant was that **stubbing** a function invocation wouldn't affect the others in the file, but yeah it doesn't make a lot of sense. So we should correct this to illustrate the actual behavior, that is any other mocks on the function's declaring class are canceled. This can be controlled if the mock is explicitly canceled after its usage, for example, once all the tests in a class execute.

PR #1067, opened 2023-03-17
Amend "Top Level functions"

polarene:
Corrected the "Top Level functions" section, to clarify the cancellation mechanism on the enclosing file.

Fixes #1061


polarene:
@mcclurej06 thanks for pointing out this mistake, could you please check if this change is suitable, or propose corrections?

Raibaz:
LGTM, @mcclurej06 if it's ok with you I will merge it.

Issue #243, opened 2019-01-31
SLF4J warnings


GrahamBorland:
I just tried MoccK as a replacement for Mockito with my Android unit tests.

```
testImplementation "io.mockk:mockk:1.9"
```
My tests pass, but I now get this warning every time I run them:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
How can I get this warning to go away?

oleksiyp:
Either exclude slf4j or provide implementation alike logback

GrahamBorland:
I was able to make the warnings go away by adding this dependency:
```
    testImplementation "org.slf4j:slf4j-nop:1.7.21"
```

oleksiyp:
Released [1.9.1](https://github.com/mockk/mockk/releases/tag/1.9.1). Please try it out and close the ticket if everything is fine.

GrahamBorland:
Works nicely now, thanks!

jd565:
@oleksiyp These warnings seem to have reappeared with 1.13.4 (probably to do with the issue above). Do we need to add the nop implementation now or is there a fix that can be put back in?

Danil0v3s:
`1.13.5` also seems to have this issue

ianbrandt:
I just updated to 1.13.7 with the hope that PR #1091 would alleviate this issue, but I'm still getting the warnings whenever there's not a SLF4J provider on the classpath.

IonutNegru87:
I've tried 1.13.8 and it seems this issue was fixed!

carterhudson:
Still seeing it in 1.13.9

mbalajee:
Seeing it in 1.13.12

PR #1091, opened 2023-05-04
Make slf4j an optional dependency

ansman:
Currently, mockk pulls in SL4J even though it's an optional dependency. This causes a warning as detailed in https://github.com/mockk/mockk/issues/243

With this change, SL4J is changed to a compileOnly dependency which truly makes it optional without having any additional work needed.

### DIFF
--- a/modules/mockk/build.gradle.kts
+++ b/modules/mockk/build.gradle.kts
@@ -38,7 +38,7 @@ kotlin {
         }
         val jvmMain by getting {
             dependencies {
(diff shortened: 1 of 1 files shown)


Raibaz:
Thanks for looking into this!

ansman:
I think the code that looks for SL4J should also change perhaps. We have several dependencies that pull in SJ4J-api and that causes mockk to think it should use it which leads to the warning in #243 
This was previously fixed but it resurfaced in #990

PR #1099, opened 2023-06-15
Issue 954: @MockKExtension.CheckUnnecessaryStub annotation is applied…

bboeck:


Raibaz:
Thanks for looking into this!

PR #1104, opened 2023-06-23
chore: converting functions to expression bodies

casplattel:
Converting some functions to expression bodies while boyscouting the code


Raibaz:
Thanks!

Issue #758, opened 2021-11-25
`@InjectMockKs` can't be used with constructor injection


volkert-fastned:
Hi. Not sure whether to report this as a bug or as a feature request.

I did some searching to see if this issue has already been discussed or raised somewhere, but I couldn't find anything.

Take the following example test class:

```kotlin
@ExtendWith(MockKExtension::class)
internal class SubjectTest(
    @MockK private val xServiceMock: XService,
    @MockK private val yServiceMock: YService,
) {
    @InjectMockKs
    private lateinit var subject: Subject

    // Test functions here.
}
```

To avoid having to use the Keyword of Shame, I prefer to use constructor injection for the Class Under Test (`Subject`) as well, so I try to rewrite it as follows:

```kotlin
@ExtendWith(MockKExtension::class)
internal class SubjectTest(
    @MockK private val xServiceMock: XService,
    @MockK private val yServiceMock: YService,
    @InjectMockKs private val subject: Subject,
) {
    // Test functions here.
}
```

I would expect this to work the same, but unfortunately this results in the following error:

```
No ParameterResolver registered for parameter [subject] in constructor...
```

I've tried this with MockK versions `1.12.0` and `1.12.1`, and it fails the same way in both.

It would be nice if `@InjectMockKs` would work through constuctor injection as well. In my opinion, `lateinit var` is ugly, and I prefer not to use it unless absolutely necessary.

Is there a specific reason why this doesn't work, like with the `@MockK` annotations? Or is this a bug?

Thanks.

Raibaz:
This is indeed an unsupported behavior; [MockKExtension](https://github.com/mockk/mockk/blob/284da25e0f6df8b291b356cde14980a8a4529073/mockk/jvm/src/main/kotlin/io/mockk/junit5/MockKExtension.kt) does not take into account parameters annotated with `@InjectMocks`.

It should be possible, though, to handle them in the `resolveParameters` method and inject the mocks; not sure if parameter order is going to be relevant here (it likely is), so we may need to force the `@InjectMocks` parameter to be the last one, but it shouldn't be a big deal.

volkert-fastned:
Such a limitation would be fine IMO, since it makes the most sense to have it as the last parameter. However, if feasible, it might be a good idea to check if that is indeed the case, and fail the test with a clear error message like `@InjectMocks is only allowed on the last constructor parameter`.

Thanks for being open to this enhancement!

volkert-fastned:
[...]

PR #1106, opened 2023-06-26
Inject mocks via constructor to avoid lateinit var

kkurczewski:
One limitation is that constructor field order is important, mock dependencies must go before actual class. In order to remove this limitation we have to ask Junit team for helping hand.

Closes https://github.com/mockk/mockk/issues/758


Raibaz (approved):

Raibaz:

PR #1108, opened 2023-07-01
Issue 963: Unnecessary stubbings detected - false positive

bboeck:


Raibaz:
Nice and clean, thanks!

jyotmansingh-tomo:
I had a few `internal class` in my Test class and this change caused my tests to start failing suddenly. I had to use `@RequireParallelTesting` as my tests were running in parallel and each internal class would clear other class's mocks. Just putting here if it helps someone else.

Issue #293, opened 2019-05-06
Bug: CapturingSlot is not working for nullable types


stengvac:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer (found captureNullable function)
- [x] I checked to make sure that this issue has not already been filed

### Current Behavior
Hello I am not sure whether following behavior is by design or not, but why  
``
class CapturingSlot<T : Any>()
`` require not null type? I have found, that I can capture nullable types with fun ``captureNullable(list)`` but can't with ``CapturingSlot``. So when I change captured type I have to use ``captureNullable(list)`` instead of ``capture(slot)``? I did not found definite answer in doc/google. 

Example

```
@ExtendWith(MockKExtension::class)
class SomeTest {
    val strSlot = slot<String>()
    
    @Test
    fun `no answer found for null`() {
        val anotherServiceMocked = mockk<AnotherService>() {
            every { callMe(capture(strSlot)) } just Runs
        }
        
        Tested(anotherServiceMocked).testedFnc(null)
    }

    @Test
    fun `answer found for not null`() {
        val anotherServiceMocked = mockk<AnotherService>() {
            every { callMe(capture(strSlot)) } just Runs
        }

        Tested(anotherServiceMocked).testedFnc("")
    }
}

class Tested(
    private val anotherService: AnotherService
) {
    fun testedFnc(str: String?) = anotherService.callMe(str)
}

class AnotherService {
    fun callMe(str: String?) = Unit
}
```

### Expected behavior
``
class CapturingSlot<T : Any?>()
`` 
so ``val stringSlot = slot<String?>()`` is possible to create and use to capture values. And capture will behave same for for null and not null values

### SW
MockK version: 1.9.3
OS: Win 10
Kotlin version: 1.3.31
JDK version: 11.0.3
jUnit version: 5.1.1

V

Jofairden:
I think I'm having the same issue atm. Trying to capture a 'JsonObject', which is marked `@Nullable`
I will actually get the error:
```
io.mockk.MockKException: 
Failed matching mocking signature for

left matchers: [slotCapture<JsonObject>()]
```
Which is just plain wrong, because the signature matches just fine. I tried captureNullable, but it only takes a mutable list. So now I also wonder why I can't capture this object, it makes me unable to complete my unit test.

Versions
MockK version: 1.9.3
OS: Win 7
Kotlin version: 1.3.30
[...]

PR #1109, opened 2023-07-04
Issue 293: support capture nullable with slot

stengvac:
Hello,

This is PR for Allow to Capture nullable values with `CapturingSlot` see [293](https://github.com/mockk/mockk/issues/293). Seems it works according to tests.

Changes on  `CapturingSlot` API are big, I can make them less invasive, but currently I don`t know how much problem it is. I mean removal of 
```
public field captured Ljava/lang/Object; //is not field only get/set now
public final fun setCaptured (Z)V // only getter now
public final fun setNull (Z)V // only getter now
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -586,41 +586,59 @@ val addressBook = mockk<AddressBook> {
 
 ### Capturing
 
-You can capture an argument to a `CapturingSlot` or `MutableList`:
+You can capture an argument to a `CapturingSlot` or `MutableList`. 
+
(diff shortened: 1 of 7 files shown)


Raibaz:
Thanks a lot for this!

Can you also please add mention of this in the docs in README.md?

stengvac:
Hello, README.md updated and added one more test for CapturingSlot.

Raibaz (approved):

PR #1117, opened 2023-07-12
Improve log message

marcelstoer:
The array.toString() implementation doesn't produce anything useful here.


Raibaz:
Nice one, thanks for looking into this!

PR #1125, opened 2023-07-25
Fix grammar in Ukraine support message

BirdHowl:
**Changes**
- Edit support message for Ukraine to fit with the intended meaning 🙂


Jolley71717 (approved):
Excellent. I believe this captures the intended meaning. The previous description was essentially the opposite of what I think mockk intended

Raibaz:
Oh, you're right, I never noticed this.

Thanks for spotting it and fixing it!

PR #1134, opened 2023-08-18
Minor performance improvements to MockInjector

k163377:
SSIA

Please see the commit comments for detailed changes.

### DIFF
--- a/modules/mockk/src/jvmMain/kotlin/io/mockk/impl/annotations/MockInjector.kt
+++ b/modules/mockk/src/jvmMain/kotlin/io/mockk/impl/annotations/MockInjector.kt
@@ -15,6 +15,13 @@ class MockInjector(
     val injectImmutable: Boolean,
     val overrideValues: Boolean
 ) {
+    companion object {
+        private val sortCriteria = compareBy<KFunction<Any>>(
+            { -it.parameters.size },
+            { fn -> fn.parameters.joinToString(",") { it.type.toString() } }
+        )
+    }
+
     fun constructorInjection(type: KClass<*>): Any {
         val firstMatching = findMatchingConstructor(type)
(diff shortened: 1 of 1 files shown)


Raibaz:
I like the usage of `fold`, thanks!

I don't really understand the reasoning behind moving `sortCriteria` to the companion object though, since it's being used only in a single place - why did you move it?

k163377:
Because `Function1[]` was being instantiated on every call.
I have not checked how much this affects performance, but I believe it is an improvement.

```kotlin
import kotlin.reflect.KFunction

val tmp = compareBy<KFunction<Any>>(
    { -it.parameters.size },
    { fn -> fn.parameters.joinToString(",") { it.type.toString() } }
)
```

```java
// decompiled
public final class TempKt {
   @NotNull
   private static final Comparator tmp;

   static {
      // `Function1[]` is generated for each call.
      tmp = ComparisonsKt.compareBy(new Function1[]{null.INSTANCE, null.INSTANCE});
   }
}
```

modules/mockk/src/jvmMain/kotlin/io/mockk/impl/annotations/MockInjector.kt
```diff
@@ -15,6 +15,13 @@ class MockInjector(
     val injectImmutable: Boolean,
     val overrideValues: Boolean
 ) {
+    companion object {
```
k163377, line 18:
```suggestion
    private companion object {
```

Raibaz:
Can you please rebase on top of the `master` branch? I mistakenly broke the build in a previous commit :(

I fixed it now

k163377:
Rebased

Raibaz:
Thanks a lot for putting this together and following up!

PR #1143, opened 2023-09-08
Reduce the use of spread operator to improve performance

### LINKED ISSUE #1139: Reduce the use of spread operator to improve performance.
I found some code that takes `vararg` as an argument and uses the spread operator to call another function that takes `vararg` as an argument.
In particular, the code that calls `io.mockk.MockKDsl`.

Spread operator is expensive operation, such as generating a copy of an array.
Therefore, it is better to use `List` to communicate internally to improve performance.

If this proposal looks good, I would like to submit a PR regarding modifications to `internal` functions.

## Benchmark score
A simple benchmark comparison shows an improvement of just under 1.2x in throughput.

```
Benchmark          Mode  Cnt        Score        Error  Units
Benchmark.list    thrpt    4  4789181.882 ± 765260.680  ops/s
Benchmark.vararg  thrpt    4  4035049.285 ± 543861.224  ops/s
```

The benchmarks performed are as follows

```kotlin
import org.openjdk.jmh.annotations.Benchmark
import org.openjdk.jmh.annotations.Level
import org.openjdk.jmh.annotations.Scope
import org.openjdk.jmh.annotations.Setup
import org.openjdk.jmh.annotations.State
import kotlin.random.Random

@State(Scope.Benchmark)
open class Benchmark {
    lateinit var arr: Array<String>

    @Setup(Level.Trial)
    fun setUp() {
        arr = Array(5) { Random.nextInt(10).toString() }
    }

    private inline fun _vararg(vararg args: String) = args.joinToString(separator = ", ")

    private inline fun _list(args: List<String>) = args.joinToString(separator = ", ")

    @Benchmark
    fun vararg() = _vararg(*arr)

    @Benchmark
    fun list() = _list(arr.asList())
}
```

https://github.com/k163377/vararg-benchmark

k163377:
Resolves #1139

### DIFF
--- a/modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
+++ b/modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
@@ -412,7 +412,7 @@ object MockKDsl {
         objects.forEach {
             val cancellation = factory.objectMockk(it, recordPrivateCalls)
 
-            internalClearMocks(it)
+            internalClearMocks(it, emptyArray())
 
             MockKCancellationRegistry
                 .subRegistry(MockKCancellationRegistry.Type.OBJECT)
@@ -466,7 +466,7 @@ object MockKDsl {
         classes.forEach {
             val cancellation = factory.staticMockk(it)
 
-            internalClearStaticMockk(it)
+            internalClearStaticMockk(arrayOf(it))
 
             MockKCancellationRegistry
                 .subRegistry(MockKCancellationRegistry.Type.STATIC)
@@ -523,7 +523,7 @@ object MockKDsl {
         classes.forEach {
             val cancellation = factory.constructorMockk(it, recordPrivateCalls, localToThread)
 
-            internalClearConstructorMockk(it)
+            internalClearConstructorMockk(arrayOf(it))
 
             MockKCancellationRegistry
                 .subRegistry(MockKCancellationRegistry.Type.CONSTRUCTOR)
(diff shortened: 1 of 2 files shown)


modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
```diff
@@ -460,13 +460,13 @@ object MockKDsl {
...
      */
-    inline fun internalMockkStatic(vararg classes: KClass<*>) {
+    inline fun internalMockkStatic(classes: Array<out KClass<*>>) {
         val factory = MockKGateway.implementation().staticMockFactory
 
         classes.forEach {
             val cancellation = factory.staticMockk(it)
 
-            internalClearStaticMockk(it)
+            internalClearStaticMockk(arrayOf(it))
```
k163377, line 469: It seems more efficient to call `internalClearStaticMockk` together as follows, but why not?
If there are no problems, I will reflect the changes.

```kotlin
classes.forEach {
    val cancellation = factory.constructorMockk(it, recordPrivateCalls, localToThread)

    MockKCancellationRegistry
        .subRegistry(MockKCancellationRegistry.Type.CONSTRUCTOR)
        .cancelPut(it, cancellation)
}

internalClearConstructorMockk(classes)
```

modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
```diff
@@ -26,7 +26,7 @@ object MockKDsl {
     inline fun <reified T : Any> internalMockk(
         name: String? = null,
         relaxed: Boolean = false,
-        vararg moreInterfaces: KClass<*>,
+        moreInterfaces: Array<out KClass<*>>,
```
k163377, line 29: I was not sure whether to use `List` or `Array`, but I chose `Array` for the following reasons.

- `MockKGateway` uses `Array` as an argument in the same situation.
- It is more efficient not to convert to `List`.
- It is an `internal` API, and it seems acceptable to sacrifice some readability.

Raibaz:
Looks good to me as it is. Thanks!

PR #1146, opened 2023-09-11
Update Android test targets

LeonRa:
## Summary

As suggested [here](https://github.com/mockk/mockk/pull/1145#issuecomment-1713924755), this updates MockK's Android test targets to be API 29 and 31 (removing 21 and adding 31). The older target is broken and no longer mainstream, while the newer is relatively common.

## Verification

- [X] Tests pass locally against an API 29 emulator (when [orchestrator fix](https://github.com/mockk/mockk/pull/1145/commits/1a1e71772f5077a68a28307c47be190c216ca03e) is applied)
- [X] Tests pass locally against an API 31 emulator (when [orchestrator fix](https://github.com/mockk/mockk/pull/1145/commits/1a1e71772f5077a68a28307c47be190c216ca03e) is applied)

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -46,7 +46,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        api-level: [ 21, 29 ] # The minSdk and targetSdk versions set for MockK Android projects
+        api-level: [ 29, 31 ] # The minSdk and targetSdk versions set for MockK Android projects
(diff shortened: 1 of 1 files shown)


LeonRa:
Looking into the failure. Need to figure out what else needs to be configured for the emulator setup

Raibaz:
Whops, merged by mistake 😓 

I think we can continue working on fixing the build on #1145 though

PR #1105, opened 2023-06-25
Update Gradle and Android dependencies

SimonMarquis:
- Gradle 8.3
- AGP 8.1.1
- targetSdk & compileSdk 34

Both updates must be applied at the same time, as they rely on each-other APIs.

PR #1148, opened 2023-09-12
Update Kotlin 1.9.10

SimonMarquis:
This will need to be rebased on top of https://github.com/mockk/mockk/pull/1105


SimonMarquis:
I've forced-push after a rebase because the merge commit re-introduced java 11 as a target but this is no longer supported by Gradle.

SimonMarquis:
I had some older commits prepared to update kotlinx-metadata-jvm some time ago, let me find them 🔍

Raibaz:
Ah sorry about reintroducing Java 11, my bad

PR #1152, opened 2023-09-12
Update KotlinX Kover 0.7.3

SimonMarquis:
https://github.com/Kotlin/kotlinx-kover/releases


SimonMarquis:
Wow, it seems like they did not publish any of the following versions on Maven Central 🤔 
https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kover/

PR #1154, opened 2023-09-12
Fix Kotlin version override in tests

### LINKED ISSUE #1048: Kotlin version configured from GitHub Actions workflow is not correctly applied

SimonMarquis:
Closes #1048

### DIFF
--- a/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
+++ b/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
@@ -15,7 +15,6 @@ object Deps {
         const val slfj = "2.0.5"
         const val logback = "1.4.5"
         const val junitJupiter = "5.8.2"
-        const val junitVintage = "5.8.2"
         const val junit4 = "4.13.2"
 
         const val byteBuddy = "1.14.6"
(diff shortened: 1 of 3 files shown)


Raibaz:
Brilliant, this is exactly what I had in mind!

Thanks a lot!

PR #1161, opened 2023-09-22
JVM 1.8 source compatibility

jeffdgr8:
[This commit](https://github.com/mockk/mockk/commit/b6cf9551da37b17f718f44677107ea23ec616982) changed JVM source compatibility to JVM 11, but this doesn't seem to be required. Reverting to 1.8 compatibility supports more apps and libraries that target JVM <11.

### DIFF
--- a/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
+++ b/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
@@ -9,7 +9,7 @@ fun Project.kotlinVersion(): String =
 
 object Deps {
     object Versions {
-        val jvmTarget = JavaVersion.VERSION_11
+        val jvmTarget = JavaVersion.VERSION_1_8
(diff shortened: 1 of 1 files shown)


merlinstardust:
How do I fix this? In my gradle file, I have `testImplementation 'io.mockk:mockk:1.13.8'` which is the latest version, but it's not working

RicardoRSousa:
> How do I fix this? In my gradle file, I have `testImplementation 'io.mockk:mockk:1.13.8'` which is the latest version, but it's not working
> 
> I'm getting the error `Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option`

This wasn't released yet. Probably will be included on `1.13.9`

merlinpaypal:
When is the 1.13.9 release expected to be published?

PR #1163, opened 2023-09-23
Remove KAPT from Android libraries

SimonMarquis:
It seems like KAPT was never actually used in this project. Building a release with and without KAPT leads to identical binaries.


SimonMarquis:
Emulator timeout, can we rerun [this job](https://github.com/mockk/mockk/actions/runs/6283263113/job/17063550001?pr=1163)?

Issue #1180, opened 2023-11-01
Warning: Namespace 'io.mockk' is used in multiple modules and/or libraries: io.mockk:mockk-android:1.13.8, io.mockk:mockk-agent-android:1.13.8.


kubode:
### Prerequisites

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

No warnings.

### Current Behavior

The following warning is logged to the console.

```
[io.mockk:mockk-android:1.13.8] ~/.gradle/caches/transforms-3/2b2aee518abef2214de7058922a1a5d2/transformed/mockk-android-1.13.8/AndroidManifest.xml Warning:
        Namespace 'io.mockk' is used in multiple modules and/or libraries: io.mockk:mockk-android:1.13.8, io.mockk:mockk-agent-android:1.13.8. Please ensure that all modules and libraries have a unique namespace. For more information, See https://developer.android.com/studio/build/configure-app-module#set-namespace
```

### Failure Information (for bugs)

The `android.namespace` config needs to be unique.

#### Steps to Reproduce

Just build with `mockk-android`.

#### Minimal reproducible code (the gist of this issue)

```kotlin
dependencies {
    testImplementation("io.mockk:mockk-android:1.13.8")
}
```

kubode:
I'm going to make a PR.

PR #1181, opened 2023-11-01
Specify unique `android.namespace` to resolve warnings

kubode:
To fix https://github.com/mockk/mockk/issues/1180

The breaking change would not occur because MockK does not use `R` files.

If you have an appropriate namespace specification, please comment.

### DIFF
--- a/modules/mockk-agent-android/build.gradle.kts
+++ b/modules/mockk-agent-android/build.gradle.kts
@@ -14,6 +14,7 @@ val mavenDescription: String by extra("${project.description}")
 
 @Suppress("UnstableApiUsage")
 android {
+    namespace = "io.mockk.proxy.android"
     externalNativeBuild {
         cmake {
             path = file("CMakeLists.txt")
--- a/buildSrc/src/main/kotlin/buildsrc/convention/android-application.gradle.kts
+++ b/buildSrc/src/main/kotlin/buildsrc/convention/android-application.gradle.kts
@@ -13,7 +13,6 @@ plugins {
 }
 
 android {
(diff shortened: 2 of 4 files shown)


modules/mockk-agent-android/build.gradle.kts
```diff
@@ -14,6 +14,7 @@ val mavenDescription: String by extra("${project.description}")
 
 @Suppress("UnstableApiUsage")
 android {
+    namespace = "io.mockk.proxy.android"
```
Raibaz, line 17: Nit: is there a reason why this is not "io.mockk.agent.android" like the module name?

kubode: Because the namespace of `mockk-agent-android-dispatcher` is `io.mockk.proxy.android.dispatcher`.
I aligned with it.
https://github.com/mockk/mockk/blob/ff79a07e3364b64fa1000d7967861d635f643395/modules/mockk-agent-android-dispatcher/build.gradle.kts#L11

kubode: Should I match the `namespace` of both this module and the `mockk-agent-android-dispatcher` module to the name of the module?

Raibaz: Nah, it's good as it is.

Thanks!

ansman:
@Raibaz Can we get a new patch release that fix this issue?

Raibaz:
I actually just released 1.13.9 :)

ansman:
I saw, nice timing. Upgraded and it works great. Thanks!

Issue #366, opened 2019-10-07
Calling method on spy of class with generic base class can cause stackoverflow


FredyH:
### Failure Information (for bugs)
Currently creating spies of classes that inherit from base classes with type parameters can cause stack overflow errors. 

- The error only happens if the method in the base class has a type parameter. When using a concrete type for the parameter instead, no error happens (even though the base class has a type parameter). 
- Furthermore this error only happens when a super call happens in the overriden method in the child class.
- It seems like none of the prints are ever executed, so the first super call most likely immediately triggers the stackoverflow.
- A similar (probably same) error has already been reported in #205 but I am creating a new issue (with hopefully more information), since the other one has been closed.

Considering generics are involved here I suspect this might have something to do with type erasure.

#### Minimal reproducible code (the gist of this issue)

```kotlin
import io.mockk.spyk

abstract class BaseClass<T> {
    open fun method(t: T) {
        println("base class $t")
    }
}

class ChildClass: BaseClass<String>() {
    override fun method(t: String) {
        super.method(t)
        println("child class $t")
    }
}

fun main() {
    val c = spyk(ChildClass())
    c.method("test")
}
```
#### Context

* MockK version: 1.9.3 (Kotlin 1.3)
* OS: Windows
* Kotlin version: 1.3
* JDK version: This error occurs on jdk 8 and 12

#### Stack trace

```
Exception in thread "main" java.lang.StackOverflowError
[....](Removed lines that are irrelevant for this bug, it all just repeats further up than this)
	at BaseClass.method(MWE.kt:8)
	at ChildClass.method(MWE.kt:13)
	at sun.reflect.GeneratedMethodAccessor6.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at io.mockk.proxy.jvm.advice.MethodCall.call(MethodCall.kt:14)
	at io.mockk.proxy.jvm.advice.SelfCallEliminatorCallable.call(SelfCallEliminatorCallable.kt:14)
	at io.mockk.impl.instantiation.JvmMockFactoryHelper.handleOriginalCall(JvmMockFactoryHelper.kt:83)
	at io.mockk.impl.instantiation.JvmMockFactoryHelper.access$handleOriginalCall(JvmMockFactoryHelper.kt:20)
[...]

PR #1195, opened 2023-12-28
Fix StackOverflowError calling method on spy of class with generic base class

flapenna:
Issue #366 

Modified the `checkOverride` function to include a check for parameter type assignability, preventing infinite recursion and resolving the stack overflow issue.


Raibaz:
Nice one, thanks a lot for looking into this!

I'll merge it as soon as the builds pass.

Issue #64, opened 2018-04-23
Bug: callOriginal() for default interface methods is not working


oleksiyp:
```
import io.mockk.every
import io.mockk.mockk
import io.mockk.spyk
import io.mockk.verify
import javax.xml.crypto.Data

data class ItemData(val id: Int)

interface Database {
    fun insertItems(items: List<ItemData>)
    fun deleteItems(items: List<ItemData>)
    fun runTransaction(alsoDelete: Boolean) {
        insertItems(listOf(ItemData(id = 1), ItemData(id = 2), ItemData(id = 3)))
        if (alsoDelete)
            deleteItems(listOf(ItemData(id = 4), ItemData(id = 5), ItemData(id = 6)))
    }
}


fun main(args: Array<String>) {
    val x = mockk<Database>()
    every { x.runTransaction(any()) } answers { callOriginal() }
    x.runTransaction(alsoDelete = true)
    verify { x.insertItems(listOf(ItemData(id = 1), ItemData(id = 2), ItemData(id = 3))) }
}
```

Should work

fargus9:
This is blocking some of my tests as well...

jsingle:
Same here

fargus9:
#important

munnitz:
What's the status on this one? I tried using this when mocking an interface like so
```kotlin
class InterfaceMock : Interface by mockk() {
    init {
        coEvery { defaultSuspendFunctionCall(any()) } coAnswers { callOriginal() }
    }
}
```
but it does not seem to work. Am I using it incorrectly or is it still not fixed?

juniofernandes-hotmart:
Same issue here. Any feedbacks?

dugsmith:
I'm also blocked writing a test like this because of this problem. Any plans to fix it?

aivinog1:
Hello everyone! I was encountered with this too and I wish to fix it :) So, I can start to fix it but are there any clues or advice about it?

zakhenry:
quick update, a much cleaner option works with abstract classes:

```kt

import io.mockk.*

data class ItemData(val id: Int)

interface Database {
    fun insertItems(items: List<ItemData>)
    fun deleteItems(items: List<ItemData>)
    fun runTransaction(alsoDelete: Boolean) {
        insertItems(listOf(ItemData(id = 1), ItemData(id = 2), ItemData(id = 3)))
        if (alsoDelete)
[...]

PR #1203, opened 2024-01-12
Fix callOriginal on default interface method

flapenna:
Issue: #64

The Kotlin compiler generates an additional class that contains an implementation of a default method as a static member. This method will always be intercepted by InterceptNoSuper since it doesn't have a `@SuperCall`.

Reference: 
https://github.com/mockk/mockk/blob/ef56926bbbc740cb0a5ae87a56726c3ca175713e/modules/mockk-agent/src/jvmMain/java/io/mockk/proxy/jvm/advice/jvm/JvmMockKProxyInterceptor.java#L34-L38

To determine whether the method is a default interface method or not, I have implemented a new utility class named DefaultInterfaceMethodResolver. 
This class attempts to find the DefaultImpls class and then the default method.


Raibaz:
Wow, thanks for fixing an issue that has been there almost forever!

PR #1204, opened 2024-01-12
Only unbox value class answers when appropriate

jselbo:
Fixes issue: https://github.com/mockk/mockk/issues/1037

Without this change, the two new test cases fail with:
```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class io.mockk.it.ValueClassTest$Companion$DummyValue (java.lang.Integer is in module java.base of loader 'bootstrap'; io.mockk.it.ValueClassTest$Companion$DummyValue is in unnamed module of loader 'app')
```


Raibaz:
Nice one, thanks for putting this together, I like the implementation!

Can you please run `gradle mockk-core:apiDump` to update the api definition files?

PR #1206, opened 2024-01-15
Make MockKExtension open to allow custom extensions

IgeNiaI:


Raibaz:
I'm not sure of what use case this would support.

Do you have an example?

IgeNiaI:
I have this utils object, note that date format is lazily initialized, and it takes current device time zone.
`
object DateTimeUtils {
    private val dateFormat by lazy {
        SimpleDateFormat("yyyy-MM-dd HH:mm:ss", Locale.getDefault()).apply {
            timeZone = TimeZone.getDefault()
        }
    }

    @Synchronized
    fun Date.asDateString(): String = dateFormat.format(this)
}`

I have tests for this object and I can mock TimeZone.getDefault() to always return same value and thus make the tests resistant to time zone changes. But the Date.asDateString() method can be called from other app components that are executed by other tests. And because of that the SimpleDateFormat object initalizes with default time zone, and then the tests will fail.

I want to mock TimeZone.getDefault() before all tests are started to make them all consistent.

IgeNiaI:
For some reason it didn't came to me that I should just mock DateTimeUtils in all other tests, even though I did that for other classes

PR #1209, opened 2024-01-18
Fix capture using mutable list of value class

ath0s:
### The Problem
When using `capture` with a mutable list of a value type (i.e. using `CaptureMatcher`), the wrapped value is captured rather than the value type:
```
expected: <DummyValue(value=101)> but was: <101>
```

_When using `capture` with a `slot` (i.e. using `CapturingSlotMatcher`) this is handled by calling the internal `boxCast` helper function._

### The Solution
Use `boxCast` in `CaptureMatcher `, in the same way as `CapturingSlotMatcher` does.


Raibaz:
Nice one, thanks for catching this and taking the time to look into it!

PR #1218, opened 2024-02-11
Embedding springmockk 4.0.2 into mockk

oleksiyp:


modules/springmockk/README.md
kkurczewski, line 1: We should also update root README as it points to old repository.

kkurczewski (commented):
So I ran test on repository which uses:
* Java 17
* Kotlin 1.8.21
* SpringMockk 4.0.2

I swapped artifact from maven central to:
1. local JAR from NinjaSquad repo before my fork
2. my fork of NinjaSquad repo after swapping packages
3. local JAR from springmockk module from this branch

First and second attempt passed while third failed with:
```
        Caused by:
        java.lang.IllegalStateException: Error processing condition on org.springframework.boot.autoconfigure.transaction.TransactionAutoConfiguration$EnableTransactionManagementConfiguration
            at org.springframework.boot.autoconfigure.condition.SpringBootCondition.matches(SpringBootCondition.java:60)
            at org.springframework.context.annotation.ConditionEvaluator.shouldSkip(ConditionEvaluator.java:108)
            at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader$TrackedConditionEvaluator.shouldSkip(ConfigurationClassBeanDefinitionReader.java:470)
            at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader$TrackedConditionEvaluator.shouldSkip(ConfigurationClassBeanDefinitionReader.java:459)
            at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader$TrackedConditionEvaluator.shouldSkip(ConfigurationClassBeanDefinitionReader.java:459)
            at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader.loadBeanDefinitionsForConfigurationClass(ConfigurationClassBeanDefinitionReader.java:131)
            at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader.loadBeanDefinitions(ConfigurationClassBeanDefinitionReader.java:120)
            at org.springframework.context.annotation.ConfigurationClassPostProcessor.processConfigBeanDefinitions(ConfigurationClassPostProcessor.java:428)
            at org.springframework.context.annotation.ConfigurationClassPostProcessor.postProcessBeanDefinitionRegistry(ConfigurationClassPostProcessor.java:289)
            at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanDefinitionRegistryPostProcessors(PostProcessorRegistrationDelegate.java:349)
            at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors(PostProcessorRegistrationDelegate.java:118)
            at org.springframework.context.support.AbstractApplicationContext.invokeBeanFactoryPostProcessors(AbstractApplicationContext.java:788)
            at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:606)
            at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:754)
            at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:456)
            at org.springframework.boot.SpringApplication.run(SpringApplication.java:334)
            at org.springframework.boot.test.context.SpringBootContextLoader.lambda$loadContext$3(SpringBootContextLoader.java:137)
            at org.springframework.util.function.ThrowingSupplier.get(ThrowingSupplier.java:58)
            at org.springframework.util.function.ThrowingSupplier.get(ThrowingSupplier.java:46)
            at org.springframework.boot.SpringApplication.withHook(SpringApplication.java:1454)
            at org.springframework.boot.test.context.SpringBootContextLoader$ContextLoaderHook.run(SpringBootContextLoader.java:552)
            at org.springframework.boot.test.context.SpringBootContextLoader.loadContext(SpringBootContextLoader.java:137)
            at org.springframework.boot.test.context.SpringBootContextLoader.loadContext(SpringBootContextLoader.java:108)
            at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContextInternal(DefaultCacheAwareContextLoaderDelegate.java:225)
            at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:152)
            ... 81 more

            Caused by:
            java.lang.IllegalArgumentException: Invalid value type for attribute 'factoryBeanObjectType': java.lang.String

```

So something was broken during migration into mockk module.

modules/springmockk/src/main/kotlin/io/mockk/springmockk/MockkClear.kt
```diff
@@ -0,0 +1,67 @@
...
+        private data class MockkClearEntry(
+            val mockRef: WeakReference<Any>,
+            var clearMode: MockkClear
+        )
+
+        // An identity hashmap would be more efficient and was used before.
+        // But it would retain references to mocks that aren't used anymore (because the Spring context cache
+        // has a limit on the number of cached contexts). See https://github.com/Ninja-Squad/springmockk/issues/97
+        // A weak hashmap would be ideal, but it uses equals and hashCode, which would cause hashCode() to be called on the mocks,
+        // and confirmVerified calls to fail. See https://github.com/Ninja-Squad/springmockk/issues/27
```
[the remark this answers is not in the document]
oleksiyp: Issues and original repo shouldnt be deleted, but archived

modules/springmockk/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+    }
+}
+
+dependencies {
+    api(projects.modules.mockk)
+    implementation(Deps.Libs.kotlinReflect)
+
+    implementation(Deps.Libs.springBootTest)
+    implementation(Deps.Libs.springTest)
+    implementation(Deps.Libs.springContext)
```
kkurczewski, line 40: I did further tests, dependency which break springmockk is: `org.springframework:spring-context:6.1.0` and above

Bumping Spring caused dependency mismatch. Following code works:
```
testImplementation("io.mockk:springmockk:1.13.10-SNAPSHOT") {
    exclude group: 'org.springframework.boot'
    exclude group: 'org.springframework'
}
//testImplementation 'org.springframework:spring-context:6.1.0' // this version breaks most recent version of springmockk
```

oleksiyp: This happens due to some internal type mismatches in `factoryBeanObjectType`. This looks like some internal interface in spring which is not standardized and changes between versions. I will try to mitigate this issue somehow.

ajax-tupis-r:
@oleksiyp when can we expect this PR to be merged?

driptaroop:
Hey there. its been sometime since the last post. is this change in the near future roadmap? thanks.

mepi262:
@Raibaz 
> I think there are still a few open points to be addressed before this can be merged.

Which part of the code do you specifically think should be modified?

Would you tell @oleksiyp  that.

thejeff77:
@oleksiyp and @kkurczewski it seems this is sort of stuck. Since this has been opened, Spring Boot 3+, Kotlin 2+ have been released which have a decent amount of breaking changes. See [here](https://github.com/mockk/mockk/pull/1218#discussion_r1492112380). 

I'd love to see this eventually get merged and maintained here so our organization can continue using SpringMockK

thejeff77:
@oleksiyp @kkurczewski @Raibaz @ajax-tupis-r 

I recommend we update this to use latest spring boot (perhaps 4 and up?), latest kotlin version (2+), Java 21, and we can point users at the old ninjasquad repo for older compatibility. 

This would move this forward without having to deal with dynamic resolution across many spring/kotlin versions before migrating to this repository is even off the ground.

bartekmarkowskiext:
@jnizet have just released new version of spring mockk, maybe there is chance to finish migration progess?

jnizet:
Hi.

I have indeed released a new version of SpringMockK, because the previous version of SpringMock didn't support Spring Boot 4, because I needed it for one of my projects, and because I didn't see it coming from MockK.

If MockK still plans to support Spring, I'm still more than happy to archive SpringMockK and to tell users to migrate to the Spring support in MockK. The code is there, ready for you to copy and paste it. All Java tests from Spring+Mockito have been migrated to Kotlin and MockK. Only one had to be disabled, but for good reasons.

Following up with the upstream changes done in the Spring+Mockito code should be easier than before, because the Mockito support is now in Spring Framework and not in Spring Boot, and the part consisting in "overriding" beans in tests is now decoupled from Mockito (except in the tests, unfortunately).

oleksiyp:
@jnizet @Raibaz I rebased PR. seems what I missed inititally was proper dependency management. it make sense to have various submodules for springmockk (named spring5mockk, spring6mockk, spring7mockk) linked against different versions of spring. I picked Spring 5, Spring 6 and Spring 7, which correspond to tags 3.1.2, 4.0.2, 5.0.1 of springmockk. so now after using exact version of spring mockk, I made all 3 submodules fully buildable with tests. I plan to add docs and push to full release this PR

jnizet:
@oleksiyp That is a bit weird, because before 5.0, SpringMockK was something that targeted Spring Boot, not Spring Framework: Spring Boot provided annotations (and their support) for Mockito, and I had reimplemented them using MockK.

Since Spring Boot 4.0, those annotations don't exist anymore in Spring Boot, because they have been refactored and integrated into Spring Framework itself. SpringMockK 5.0 is a rewrite based on those new annotations in Spring Framework.

And Spring Boot and Spring Framework don't use the same version numbers.

Frankly, I think MockK should only provide support for Spring starting at version 7 (and thus Spring Boot starting at version 4). For older versions of Spring Boot, let the developers keep using SpringMockK: it probably won't evolve anymore anyway and it's already well established.

oleksiyp:
@jnizet right, it might be rebound to spring boot version aka as a companion to spring-boot-test. while, I think, having support for previous  versions, which is still in active use, make sense. while naming convention might be different: for example "io.mockk:springmockk:1.14.7" for Spring 7 / Boot 4, and "io.mockk:springmockk-boot3:1.14.7" for Boot 3, and "io.mockk:springmockk-boot2:1.14.7" for Boot 2.

Indeed, I stopped being Java dev when Spring Framework 5 came, so not exactly sure how all things evolve and what is usage stats.

jnizet:
Here are my thoughts.

### Spring Boot 2.x: 

MockK shouldn't support it anymore IMHO. It's too old, and is not supported anymore by Spring (see https://endoflife.date/spring-boot)

### Spring Boot <= 3.4: 

MockK shouldn't support it anymore IMHO. It's too old, and will not have OSS support anymore in one month (see https://endoflife.date/spring-boot)

### Spring Boot 3.5: 

MockK might support it, since it's still very much used (version 4.0 just came out). That's the version where Spring Boot still had its Mockito annotations, but Spring Framework introduced its own ones, and Spring Boot asked users to migrate to the Spring Framework annotations. 

So there are several choices here:
 1. provide the Spring Boot annotations equivalent. This can be used by using the code of SpringMockK 4.2.x
 2. provide the Spring Framework annotations equivalent. The easiest way would probably be to start from the code of SpringMockK 5.0, and modify it to revert the changes that happened between Spring 6.2 and Spring 7. It's a bit cumbersome, but there aren't so many changes
 3. not provide anything, since SpringMockK already provides the support. This version of Spring Boot is not the latest one, so new projects will probably start with Spring Boot 4 or Spring framework 7. And old projects probably already use SpringMockK anyway. So there's not much benefit in supporting version 3.5, IMHO.

### Spring Boot 4.0.0: 

MockK should support it. That can be done by using the code of SpringMockK 5, as is.

### Versioning

Regarding versioning, I think having a separate project for each version of Spring (or Spring Boot if MockK supports it) is indeed probably the best way. That will make it possible to still support Spring 7 when Spring 8 comes out and introduces breaking changes.

So MockK could have

- springmockk7:1.14.7 that would support Spring 7
- in the future: springmockk8:1.17.5 that would support Spring 8
- springmockk6:1.14.7 that would support Spring 6 if MockK chooses to support it
- springmockk-boot3:1.14.7 that would support Spring Boot 3 if MockK chooses to support it

### Note on SpringMockk 5's code

To be able to port some tests of Spring to SpringMockK, I had to use reflection in 2 or 3 places, because the tests called classes or methods that were package-private. 
That makes the code of those tests a bit brittle of course: they could still compile even though the classes/methods have been renamed or refactored. So if MockK decides to support Spring, you should maybe ask Spring to make those classes more accessible to avoid reflection.

Also, Spring Boot rejected the idea of integrating MockK support directly inside Spring Boot. Now that those annotations are part of Spring Framework, and that the code specific to Mockito (and MockK) is really small because the bean overriding mechanisms have been extracted, they might want to reconsider the decision. There's no reason not to ask. It would be easier for them to introduce the same changes in both mocking frameworks that it would be for MockK.

oleksiyp:
@jnizet ChatGPT may lie and there is just not enough data over Internet, but I would internally guestimate same - that legacy systems with Boot 2.7 exists:

image

While it is enough to just pull tags of springmockk on demand for such cases

homeworkprod:
> […] I would internally guestimate same - that legacy systems with Boot 2.7 exists:

I can confirm that they do … :grimacing:

thejeff77:
Happy New Year everybody! Anything I can do to help this along? @oleksiyp @jnizet @bartekmarkowskiext @kkurczewski

expe-elenigen:
Hi, would it be possible to clarify the situation here, does mockk still plan to integrate springmockk project? It's been few years already and I see now that springmockk still has some activity now:
https://github.com/Ninja-Squad/springmockk/releases/tag/5.0.0
Thank you!

PR #518, opened 2020-10-30
Added `mockkStatic` hard reference support

gmazzo:
Adds a `jvm`-only `mockkStatic` overload function that will allow passing a function reference to pick up the `declaringClass`.

This allows to write more robust tests as hardocded class references like:
```kotlin
// declared in File.kt ("pkg" package)
fun Obj.extensionFunc() = value + 5

mockkStatic("pkg.FileKt")
```
Can now be just:
```kotlin
// declared in File.kt ("pkg" package)
fun Obj.extensionFunc() = value + 5

mockkStatic(Obj::extensionFunc)
```

PR #1239, opened 2024-04-02
Implement `clearStaticMockk` for KFunction and KProperty

krocard:
For a class, Mockk allows to:
- `mockkStatic(klass)`
- `unmockkStatic(klass)`
- `clearStaticMock(klass)`

But for static properties and functions, since #518, only the first 2 are supported:
- `mockkStatic(::staticMethod)`
- `unmockkStatic(::staticMethod)`
- `clearStaticMock(::staticMethod)` // Doesn't compile

The helper functions were missing.


krocard:
`jvmApiCheck` is failing. Probably because the PR is adding a new function. Is there a way to update the *api files automatically?

PR #1240, opened 2024-04-03
[CONTRIBUTING.md] Suggest running `gradle check` before opening a PR

krocard:
I opened a PR that didn't pass the CI (#1239) even after reading CONTRIBUTING.md:
> Make sure all tests pass and you have a passing build for the whole project.

I did run `./gradlew test` but the CI runs the superset `./gradlew check`.

How about referencing the exact command to run before submitting the PR at the end of that line:
~~~diff
- 1. Make sure all tests pass and you have a passing build for the whole project.
+ 1. Make sure all tests pass and you have a passing build for the whole project (run `./gradlew check`).
~~~

~It would also be nice to document `./gradlew apiDump` (only found it after manually fixing the API file), but keeping the instructions short is probably better. What do you think?~ It's referenced in the `./gradlew check` error message.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ Please note we have a code of conduct, please follow it in all your interactions
 
 ## Pull Request Process
 
-1. Make sure all tests pass and you have a passing build for the whole project.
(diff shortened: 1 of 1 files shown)


krocard:
Any opinion on this change?

PR #1287, opened 2024-08-17
Add anyNullable

NWuensche:
Add the `anyNullable` matcher.


Raibaz:
Nice! Can you please add test coverage?

NWuensche:
I added a test, sorry for the late response

Issue #813, opened 2022-04-14
Cannot mock function returning nullable value class


TimonKanters-TomTom:
### Expected Behavior

Being able to mock functions returning nullable `value class`es.

### Current Behavior

It's possible to mock functions returning `value class`es, but only if they return a non-null value. Otherwise an exception is thrown while trying to execute the `every` statement.

#### Steps to Reproduce

Create a class with a function that returns a nullable `value` class (e.g., `fun getInstance(): TestValueClass?`) and try to mock its `getInstance()` call.

Note that `fun getInstance(): TestValueClass` (without `?`) can be mocked successfully.

#### Context

* MockK version: 1.12.3
* OS: Windows 10
* Kotlin version: 1.6.10
* JDK version: 1.8
* JUnit version:4.13.2
* Type of test: unit test

#### Stack trace

```
io.mockk.MockKException: Class cast exception happened.
Probably type information was erased.
In this case use `hint` before call to specify exact return type of a method.

io.mockk.MockKException: Class cast exception happened.
Probably type information was erased.
In this case use `hint` before call to specify exact return type of a method.

	at app//io.mockk.impl.InternalPlatform.prettifyRecordingException(InternalPlatform.kt:74)
	at app//io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:66)
	at app//io.mockk.impl.eval.EveryBlockEvaluator.every(EveryBlockEvaluator.kt:30)
	at app//io.mockk.MockKDsl.internalEvery(API.kt:93)
	at app//io.mockk.MockKKt.every(MockK.kt:98)
	at app//<omitted>.ValueClassMockKTest.return nullable value class from mockk(ValueClassMockKTest.kt:98)
	at java.base@11.0.2/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base@11.0.2/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base@11.0.2/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base@11.0.2/java.lang.reflect.Method.invoke(Method.java:566)
	at app//org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
[...]

PR #1295, opened 2024-09-12
fix(813): Handle nullable value classes

VasilisDrettas-tomtom:
Fix https://github.com/mockk/mockk/issues/813

Problem:
When trying to mock a function that returns a nullable value class, a ClassCastException was thrown. Thats because  mockk is using the boxedClass. Nullable value classes should be handled differently because they are not inlined.

For example
`Caused by: java.lang.ClassCastException: class java.lang.Integer cannot be cast to class io.mockk.it.ValueClassTest$Companion$DummyValue (java.lang.Integer is in module java.base of loader 'bootstrap'; io.mockk.it.ValueClassTest$Companion$DummyValue is in unnamed module of loader 'app')`

Implementation:
When the return type of the function or a member property is nullable do not use the boxedClass but use the value class itself

### DIFF
--- a/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
+++ b/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
@@ -31,8 +31,13 @@ actual object ValueClassSupport {
             //   method.returnType == int (the actual representation of inlined property on JVM)
             //   method.kotlinFunction.returnType.classifier == Foo
             val expectedReturnType = kFunction.returnType.classifier
+            val isReturnNullable = kFunction.returnType.isMarkedNullable
             val isPrimitive = resultType.boxedClass.java.isPrimitive
-            return if (!(kFunction.isSuspend && isPrimitive) && resultType == expectedReturnType) {
+            return if (
+                !(kFunction.isSuspend && isPrimitive) &&
+                resultType == expectedReturnType &&
+                !isReturnNullable
+            ) {
                 this.boxedValue
             } else {
                 this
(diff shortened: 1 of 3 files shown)


Raibaz:
Thanks for looking into this!

dadadom:
I *think* this might have broken some of my testcases which worked with 1.13.12.

```kotlin
    @JvmInline
    private value class MyId(private val id: String)

    private interface Y {
        suspend fun getId(): MyId?
    }

    private class X(private val y: Y) {
        suspend fun doStuff(): MyId? = y.getId()
    }

    @Test
    fun `test stuff`() {
        val mocked = mockk<Y>()
        coEvery { mocked.getId() } returns MyId("987")

        runBlocking { X(mocked).doStuff() } shouldBe MyId("987")
    }
```
Error message:
```
12:26:43.893 [Test worker] DEBUG io.mockk.impl.instantiation.AbstractMockFactory -- Creating mockk for Y name=#1
12:26:44.895 [Test worker @coroutine#4] DEBUG io.mockk.impl.recording.states.AnsweringState -- Answering MyId(id=987) on Y(#1).getId-oOl4dpc(continuation {})

class MyTest$MyId cannot be cast to class java.lang.String (MyTest$MyId is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class MyTest$MyId cannot be cast to class java.lang.String (MyTest$MyId is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
	at MyTest$X.doStuff-oOl4dpc(MyTest.kt:17)
	at MyTest$test stuff$2.invokeSuspend(MyTest.kt:25)
...
```

VasilisDrettas-tomtom:
You are right @dadadom, it seems to be a regression due to my change. I will attempt to fix the regression once I find sometime.

Issue #820, opened 2022-05-13
Do not unmock all mocks after every test function, but after the entire test class only. Clear mocks after every test function.


erikhuizinga:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

The provided JUnit 4 rule /. JUnit 5 extension efficiently run before/after test functions and classes.

### Current Behavior

The provided JUnit 4 rule / JUnits 5 extension unmock after every test.

#### Context

In general, mocking can be expensive. Do not unmock unless the test class is completely done. Prefer clearing mocks instead of unmocking between test class functions.

The suggestion is to do the following for both the JUnit 4 rule and the JUnit 5 extension:
- Before tests, init annotations
- After every test function, clear all mocks.
- After all test functions (after a test class), unmock all mocks.

#### V

*  MockK version: 1.12.4

PR #1297, opened 2024-09-17
Run `unmockkAll` after each JUnit 5 test

erikhuizinga:
This PR originally aimed to fix https://github.com/mockk/mockk/issues/820. What it now does is:

1. Add benchmarks (results below) for `clearAllMocks()` and `unmockkAll()`, to demonstrate which of the two is more performant.
2. Make the JUnit 5 `MockKExtension` behave like the JUnit 4 `MockKRule`, making it call `unmockkAll` after each unit test function. This PR makes `MockKExtension` respect the JUnit 5 test instance lifecycle (either per method (default) or per class).
   - Note the work done in https://github.com/mockk/mockk/pull/739. That PR assumed that nobody wants to unmock after each test. However, that should probably be the default practice to prevent leaking mocks, while allowing to deviate from it.
3. Update the Gradle wrapper to the latest version, because I noticed that the checked in wrapper jar and shell scripts were outdated.
4. Update the setup-gradle action to v4, which performs Gradle wrapper validation. This is recommendable in open source repos that accept public contributions. E.g. see https://github.com/mockk/mockk/actions/runs/10904085959/job/30259684320?pr=1297#step:4:33.
5. Update the benchmark dependencies used in this PR.

---

# Benchmark results (on my local machine)

Input:

`$ ./gradlew benchmark`

Output:

```
main summary:
Benchmark                                     Mode  Cnt         Score           Error  Units
JmhTest.clearAllMocksAfterNoMockOrStub       thrpt    3  23288167.041 ± 443565082.712  ops/s
JmhTest.clearAllMocksAfterSimpleMock         thrpt    3       227.068 ±       988.568  ops/s
JmhTest.clearAllMocksAfterSimpleMockAndStub  thrpt    3       205.985 ±      1302.086  ops/s
JmhTest.noMockOrStub                         thrpt    3  79739137.777 ±   8111406.644  ops/s
JmhTest.simpleMock                           thrpt    3     62802.584 ±   1707748.668  ops/s
JmhTest.simpleMockAndStub                    thrpt    3     79292.319 ±     64931.152  ops/s
JmhTest.unmockkAllAfterNoMockOrStub          thrpt    3  39488828.944 ±  16722809.880  ops/s
JmhTest.unmockkAllAfterSimpleMock            thrpt    3     36727.737 ±   1064465.208  ops/s
JmhTest.unmockkAllAfterSimpleMockAndStub     thrpt    3     76428.656 ±     19136.524  ops/s
```

## Conclusion

Contrary to what I originally claimed in https://github.com/mockk/mockk/issues/820, `unmockkAll` performs much faster (more operations per second) in the benchmarks in this PR, except when there's nothing to unmock (then `clearAllMocks` also is fast). This means that it's probably a good idea to unmock all after each unit test. This is what `MockKRule` does if applied as a JUnit 4 `@Rule`. However, before this PR the JUnit 5 `MockKExtension` doesn't call `unmockkAll` after each test. This PR fixes that.

In this way, https://github.com/mockk/mockk/issues/820 can now also be closed (after accepting these changes), since it is outdated / incorrect.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.10.1-bin.zip
```
erikhuizinga, line 3: See http://services.gradle.org/versions/current

test-modules/performance-tests/build.gradle.kts
```diff
@@ -4,11 +4,11 @@ import kotlinx.benchmark.gradle.benchmark
 
 plugins {
     buildsrc.convention.`kotlin-jvm`
-    id("org.jetbrains.kotlinx.benchmark") version "0.4.5"
+    id("org.jetbrains.kotlinx.benchmark") version "0.4.12"
```
erikhuizinga, line 7: See https://github.com/Kotlin/kotlinx-benchmark/tags

test-modules/performance-tests/build.gradle.kts
```diff
@@ -26,7 +26,7 @@ benchmark {
     targets {
         register("main") {
             this as JvmBenchmarkTarget
-            jmhVersion = "1.35"
+            jmhVersion = "1.37"
```
erikhuizinga, line 29: See https://github.com/openjdk/jmh/tags

Raibaz:
This is awesome, thanks a lot!

PR #1305, opened 2024-10-09
Add tests for non-mockable classes

S-YED:
### Description

This pull request adds tests for non-mockable classes to ensure that attempting to mock these classes results in the expected warnings or exceptions. The tests cover commonly non-mockable classes such as `System`, `File`, and `Path`.

### Changes

- Added `NonMockableClassesTest.kt` in `modules/mockk/src/jvmTest/kotlin/io/mockk/`.

- Included test cases to verify that mocking non-mockable classes throws `IllegalArgumentException`.

### Motivation and Context

These tests are added to improve the robustness of the library by explicitly handling non-mockable classes and ensuring that appropriate exceptions are thrown when such classes are mocked.

### How Has This Been Tested?

- Ran the tests locally using `./gradlew test`.

- Verified that the tests pass and the correct exceptions are thrown.

### Related Issue

(If applicable, link to the issue here)

### DIFF
--- a/modules/mockk/src/commonMain/kotlin/io/mockk/MockK.kt
+++ b/modules/mockk/src/commonMain/kotlin/io/mockk/MockK.kt
@@ -1,8 +1,70 @@
+import kotlin.reflect.KClass
+
+
+
+private val nonMockableClasses = listOf(
+
+  System::class,
+
+  java.io.File::class,
+
+  java.nio.file.Path::class,
+
+  // Add more classes as necessary
+
+)
+
+
+
 @file:Suppress("NOTHING_TO_INLINE")
+
+
 
 package io.mockk
 
-import kotlin.reflect.KClass
+
+
+inline fun <reified T : Any> mockk(
+
+  name: String? = null,
+
+  relaxed: Boolean = false,
+
+  vararg moreInterfaces: KClass<*>,
+
+  relaxUnitFun: Boolean = false,
+
+  block: T.() -> Unit = {}
+
+): T {
+
+  if (nonMockableClasses.contains(T::class)) {
+
+    throw IllegalArgumentException("Mocking of ${T::class.simpleName} is not allowed.")
+
+  }
+
+  return MockK.useImpl {
+
+    MockKDsl.internalMockk(
+
+      name,
+
+      relaxed,
+
+      moreInterfaces,
+
+      relaxUnitFun = relaxUnitFun,
+
+      block = block
+
+    )
+
+  }
+
+}
+
+
 
 /**
  * Builds a new mock for specified class.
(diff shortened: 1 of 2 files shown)


Raibaz:
Thanks for looking into this!

However, I think throwing exceptions by default is a bit too much and will likely break too many existing applications; I think it'd be better to just log a warning.

modules/mockk/src/commonMain/kotlin/io/mockk/MockK.kt
```diff
@@ -1,8 +1,70 @@
+import kotlin.reflect.KClass
+
+
+
+private val nonMockableClasses = listOf(
+
+  System::class,
+
```
Raibaz, line 8: I think you have a code formatter that adds a ton of newlines. Can you please not add needless empty lines?

modules/mockk/src/commonMain/kotlin/io/mockk/MockK.kt
```diff
@@ -1,8 +1,70 @@
...
 @file:Suppress("NOTHING_TO_INLINE")
 
+
+
 package io.mockk
 
-import kotlin.reflect.KClass
+
+
+inline fun <reified T : Any> mockk(
```
Raibaz, line 27: Also, this is just a wrapper for the `MockKDsl` class, I don't think it should contain any business logic.

Can you please move the check to `AbstractMockFactory`?

modules/mockk/src/commonMain/kotlin/io/mockk/NonMockableClassesTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+
+
+class NonMockableClassesTest {
+
+
+
+  @Test
+
+  fun `should throw exception when mocking System class`() {
```
Raibaz, line 17: I think you can make this test parameterized and pass the list of non mockable classes.

Raibaz (changes requested):

Raibaz:
Can you also please make sure the PR builds?

modules/mockk/src/commonMain/kotlin/io/mockk/MockK.kt
S-YED: Sure i will remove em

modules/mockk/src/commonMain/kotlin/io/mockk/MockK.kt
S-YED: SURE!!!

modules/mockk/src/commonMain/kotlin/io/mockk/NonMockableClassesTest.kt
S-YED: Can you Help me with it ?

PR #1309, opened 2024-10-10
fix(1073): value class wrapped in Kotlin result is unboxed

mkotsbakws:
It should not be unboxed.


mkotsbakws:
This is strange. New test fail when I run it in IntelliJ, but not here and from command line. Maybe dependent on JVM/Kotlin version?

Issue #1308, opened 2024-10-10
[1.13.13] ClassCastException when mocking function returning a nullable value class containing another value class


felixkrull-neuland:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

Can mock a function returning a nullable value class containing another value class

### Current Behavior

ClassCastException when calling the mocked function returning aforementioned value class sandwich

### Failure Information (for bugs)

This is a regression from 1.13.12 to .13 when returning nested value classes from a mocked function. 

#### Steps to Reproduce

1. have a value class
2. have another value class containing the first value class
3. have a function returning the second value class, nullable
4. mock the function
5. call the mocked function

Removing the nullability makes the problem go away, as does moving the nullability into either of the value classes.

#### Context

I'm guessing this is related to the fix for https://github.com/mockk/mockk/issues/813

#### Stack trace

```
Exception in thread "main" java.lang.ClassCastException: class Outer cannot be cast to class java.lang.String (Outer is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
	at Cls.returnValueClass-dzOYYKU(Repro.kt:11)
	at ReproKt.main(Repro.kt:17)
	at ReproKt.main(Repro.kt)

```

#### Minimal reproducible code (the gist of this issue)

```kotlin
// --------- Gradle deps
dependencies {
    implementation("io.mockk:mockk:1.13.13")
}

// --------- code ---------
import io.mockk.every
import io.mockk.mockk

@JvmInline
value class Inner(val underlying: String)

@JvmInline
value class Outer(val inner: Inner)

open class Cls {
    fun returnValueClass(): Outer? = null
}

fun main() {
    val mock = mockk<Cls>()
    every { mock.returnValueClass() } returns Outer(Inner("mock"))
    println(mock.returnValueClass())
}
```

TWiStErRob:
Even simpler:

```
@JvmInline
value class V(val underlying: String)

interface Subject {
    fun returnValueClass(): V?
}

    @Test
    fun test() {
        val mock = mockk<Subject>()
        every { mock.returnValueClass() } returns V("data")

        val ret = mock.returnValueClass()

        assertEquals(V("data"), ret)
    }
```

Changing to non-null (= removing `?` from interface) works as expected.

NickyNOS:
[...]

PR #1314, opened 2024-10-18
fix(1308): Handle nullable complex and nested value classes

VasilisDrettas-tomtom:
Fix https://github.com/mockk/mockk/issues/1308

Problem:
The [PR](https://github.com/mockk/mockk/pull/1295) introduced a regression when a function is returning a nullable value class with a property that is not primitive and also in cases of a nested value class where the innermost property is not a primitive. In these cases we should be using the boxed value instead of the unboxed value.

Implementation:
The implementation was adjusted based on the following.
- When the innermost property of the value class is primitive then use the unboxed value
- When the innermost property of the value class is NOT primitive then use the boxed value. 
- Added some tests for nested value class as a return type of a suspending functions.
- Revert the change in JvmMockFactoryHelper.kt as it's not necessary

### DIFF
--- a/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
+++ b/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
@@ -32,11 +32,11 @@ actual object ValueClassSupport {
             //   method.kotlinFunction.returnType.classifier == Foo
             val expectedReturnType = kFunction.returnType.classifier
             val isReturnNullable = kFunction.returnType.isMarkedNullable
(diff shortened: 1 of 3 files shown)


Raibaz:
Nice, thanks for looking into this!

I'm both amazed and scared at the recursive implementation of `innermostBoxedClass`, but I think that was the only way to do this correctly.

TWiStErRob:
Wow, that looks simple yet complex, nice one! Thanks so much for looking!

matusekma:
Hi @Raibaz,
could you please make a release. We would like to update our mockk version but this bug prevents us from doing so. Thank you!

Issue #1330, opened 2024-12-16
Relaxed mocking fail to provide  simple value when property is nested value class


pbajurko:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [ x] I am running the latest version
- [ x] I checked the documentation and found no answer
- [ x] I checked to make sure that this issue has not already been filed

### Expected Behavior

A relaxed mock should return simple value for property is  nested value class.

### Current Behavior

Mocked class throws ClassCastException when property getter called.

### Failure Information (for bugs)

Test passes on 1.13.9 but not on 1.13.10, regression probably caused by #1202
Related issue #1245

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.13.13
* OS: 15.2
* Kotlin version: 1.8.22
* JDK version: 17
* JUnit version: 
* Type of test: unit test

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
class InnerClass cannot be cast to class java.lang.Long (InnerClass is in unnamed module of loader 'app'; java.lang.Long is in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class InnerClass cannot be cast to class java.lang.Long (InnerClass is in unnamed module of loader 'app'; java.lang.Long is in module java.base of loader 'bootstrap')
	at ContainerOfOuterClass.getFoo-7FQk4JE(DefaultValueinMockedValueClassTest.kt:14)
	at DefaultValueinMockedValueClassTest.testDefaultValue(DefaultValueinMockedValueClassTest.kt:29)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
[...]

PR #1331, opened 2024-12-19
Fix(Issue no. 1330) for Relaxed Mocking Value When Property is Nested Value Class

kpadhiamex:
Fix for the Issue: https://github.com/mockk/mockk/issues/1330 

Fixed for given nested value classes when relaxed mocking is enabled then value successfully verified.
Added couple of test(s) so that it will be verified in every PR.

**Root Cause**
Because of this PR's: https://github.com/mockk/mockk/pull/1314/files change in the below code in the else part, this issue was broken. Instead of returning this in the else block which returns the value class, instead check if there is any boxedValue and return the same if exists else return the same value class.

**Earlier**
```
return if (resultType == expectedReturnType && !(isReturnNullable && isPrimitive)) {
    this.boxedValue
} else {
    this
}
```

**Now**
```
return if (resultType == expectedReturnType && !(isReturnNullable && isPrimitive)) {
    this.boxedValue
 } else {
    this.boxedValue ?: this
 }
```

### DIFF
--- a/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
+++ b/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
@@ -55,7 +55,7 @@ actual object ValueClassSupport {
             return if (resultType == expectedReturnType && !(isReturnNullable && isPrimitive)) {
                 this.boxedValue
             } else {
-                this
+                this.boxedValue ?: this
             }
         }
     }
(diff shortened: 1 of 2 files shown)


kpadhiamex:
@Raibaz Can I ask, kindly review this PR if you are available?

Raibaz:
Hi @kpadhiamex, thanks for looking into this!

PR looks good to me, but to merge it I need the builds to pass. Can you please rebase on top of `master`? I just pushed a fix to the build process.

Raibaz (approved):

kpadhiamex:
Hi @Raibaz , Many thanks for your comments and approval. I just merged master branch, could you please have a look and approve the workflow which is waiting for approval?

`1 workflow awaiting approval`

Issue #1073, opened 2023-04-01
Issue with mocking value classes with coEvery


Syex:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Failure Information (for bugs)

I'm running into an edge case issue with value classes being used as mocks for suspending methods that implement an interface with generics.

Consider the following interface with an implementing class using Kotlin `Result` (or any other value class)

```
interface Action<Params, ReturnType> {

    suspend fun execute(params: Params): ReturnType
}

class ResultTest : Action<Unit, Result<String>> {

    override suspend fun execute(params: Unit): Result<String> {
        TODO("Not yet implemented")
    }
}
```

and a simple test like 

```
private val resultTest = mockk<ResultTest>()

@Test
fun test() = runTest {
    coEvery { resultTest.execute(Unit) } returns Result.success("abc")

    val result = resultTest.execute(Unit)
    assertTrue(result.isSuccess)
}
```

it fails with 

> class java.lang.String cannot be cast to class kotlin.Result

when calling `resultTest.execute(Unit)`.

---

With a modification of the return type of the interface method `execute` from `ReturnType` to `Result<ReturnType>`:

```
interface Action<Params, ReturnType> {

    suspend fun execute(params: Params): Result<ReturnType>
}

class ResultTest : Action<Unit, String> {

    override suspend fun execute(params: Unit): Result<String> {
        TODO("Not yet implemented")
    }
}
```

the same test will pass.

The test will also pass with the fist version of the interface if I remove the `suspend` keyword and replace `coEvery` with `every`.

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.13.4
* Kotlin version: 1.8.10
* JDK version: 11
* JUnit version: 5.9.2
* Type of test: unit test OR android instrumented test: Unit

#### Stack trace

```
class java.lang.String cannot be cast to class kotlin.Result (java.lang.String is in module java.base of loader 'bootstrap'; kotlin.Result is in unnamed module of loader 'app')
java.lang.ClassCastException: class java.lang.String cannot be cast to class kotlin.Result (java.lang.String is in module java.base of loader 'bootstrap'; kotlin.Result is in unnamed module of loader 'app')
[...]

Issue #859, opened 2022-07-24
Support nested value classes


aSemy:
Value class support is being improved in #849 

However there are issues when a value class is nested, for example

```kotlin
@JvmInline
value class DummyValue(val value: Int)

@JvmInline
value class DummyValueWrapper(val value: DummyValue)
```

Stubbing methods which use wrapped types causes exceptions

```none
Caused by: java.lang.ClassCastException: io.mockk.it.ValueClassNestedTest$Companion$DummyValue cannot be cast to java.lang.Integer
	at io.mockk.it.ValueClassNestedTest$Companion$DummyValue.getValue(ValueClassNestedTest.kt:471)
```

And using `any()`

```none
Failed matching mocking signature for
SignedCall(retValue=0, isRetValueMock=false, retType=class kotlin.Int, self=DummyService(#1087), method=argWrapperReturnValueClass-_ACQZD0(Int), args=[-1016248876], invocationStr=DummyService(#1087).argWrapperReturnValueClass-_ACQZD0(-1016248876))
left matchers: [any()]
io.mockk.MockKException: Failed matching mocking signature for
SignedCall(retValue=0, isRetValueMock=false, retType=class kotlin.Int, self=DummyService(#1087), method=argWrapperReturnValueClass-_ACQZD0(Int), args=[-1016248876], invocationStr=DummyService(#1087).argWrapperReturnValueClass-_ACQZD0(-1016248876))
left matchers: [any()]
```

And using `slot()`

```none
Failed matching mocking signature for
SignedCall(retValue=, isRetValueMock=true, retType=class io.mockk.it.ValueClassNestedTest$Companion$DummyValue, self=DummyService(#1086), method=argWrapperReturnWrapper-QYKvEyQ(Int), args=[376729697], invocationStr=DummyService(#1086).argWrapperReturnWrapper-QYKvEyQ(376729697))
SignedCall(retValue=0, isRetValueMock=false, retType=class kotlin.Int, self=, method=getValue(), args=[], invocationStr=DummyValue(temporary mock).getValue())
left matchers: [slotCapture<DummyValueWrapper>()]
io.mockk.MockKException: Failed matching mocking signature for
[...]

PR #1332, opened 2025-01-03
Fix( Issue #1073): Bug fix for the issue with mocking value classes with coEvery

kpadhiamex:
Fix( Issue #[1073](https://github.com/mockk/mockk/issues/1073)): Bug fix for the issue with mocking value classes with coEvery.
Enable few disabled tests for support nested value classes (issue: [859](https://github.com/mockk/mockk/issues/859))

### DIFF
--- a/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
+++ b/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
@@ -13,6 +13,8 @@ import kotlin.reflect.jvm.kotlinFunction
 
 actual object ValueClassSupport {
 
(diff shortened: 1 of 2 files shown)


kpadhiamex:
@Raibaz Could you please review this PR?

Raibaz:
Out of curiosity, why are you converting the class names to strings when doing comparisons? Does the class to class comparison not work?

kpadhiamex:
@Raibaz 
Used class comparison now for the first part, tried earlier but tried again with another way which worked however, kept `!unboxValueReturnTypes.contains(expectedReturnType.toString()) ` as is as the Result class as that doesn't seem to work. Any idea if that can be compared as well?

Raibaz (approved):

Raibaz:
Thanks for putting this together and for the follow up commits!

Issue #1329, opened 2024-12-14
MockK 1.13.13 issues with parallel testing


ebrowne72:
V1.13.13 added calling `unmockkAll()` after tests when using the JUnit 5 MockKExtension. It has the same problems with running tests in parallel that happened when the extension was modified to call `clearAllMocks()` after tests. The call to `unmockkAll()` should be wrapped in a check for `!context.requireParallelTesting`.

Raibaz:
Good catch! I'd gladly review a PR to add that check.

PR #1335, opened 2025-01-17
Fix (issue# 1329) parallel testing for unmockkAll

kpadhiamex:
Fixes https://github.com/mockk/mockk/issues/1329

V1.13.13 added calling unmockkAll() after tests when using the JUnit 5 MockKExtension. It has the same problems with running tests in parallel that happened when the extension was modified to call clearAllMocks() after tests. The call to unmockkAll() is wrapped in a check with `!context.requireParallelTesting` in this PR.


kpadhiamex:
@Raibaz Could you please review this PR?

Raibaz:
Nice, thanks for putting this together!

Do you think it'd be possible to add test coverage for it?

kpadhiamex:
@Raibaz 
Added some tests, could you please see if that is what you were looking for? If not, please suggest a reference.

Issue #1304, opened 2024-10-08
Throw a warning when mocking classes that should not be mocked


Raibaz:
There are a few classes that it is generally a bad idea to mock: trying to mock them is usually a code smell indicating that a test is not testing the correct part of the application, and mocking them usually leads to problems.

Some of these are:
- `System`
- Collections (particularly `HashMap`)
- Data classes, and in general value objects
- Classes that deal with I/O, like `File` or `Path` (you should be wrapping them in a layer that only deals with I/O rather than mixing it with business logic)

Mockk should _at least_ log a warning when trying to mock one of these classes, but ideally it should also have a configuration flag that disables mocking them altogether and has it throw an exception when a user tries to.

We could maybe even make the list of non-mockable classes configurable.

RaulErnesto08:
Hello!
I'm looking to contribute to open-source projects, can I work on this issue? 😄

Raibaz:
Sure, go ahead!

MahmoudMabrok:
Hi @Raibaz is it still valid to have next issue open when we will raise exception for `File` class

https://github.com/mockk/mockk/issues/603

PR #1340, opened 2025-01-30
[#1304] feat: Restrict mocking of certain classes and add configuration option 

devtaebong:
## Summary
This PR introduces a mechanism to prevent mocking of certain classes that are generally considered bad practice.  
By default, attempting to mock restricted classes will log a warning. Additionally, a new configuration option has  
been added to allow users to enforce strict restrictions, causing an exception to be thrown when attempting to mock  
a restricted class.

### **1️⃣ Restricted Mocking Mechanism**
- Added `RestrictedMockClasses` to maintain a list of non-mockable classes.
- By default, logs a warning when attempting to mock a restricted class.

### **2️⃣ New Configuration Option**
- Introduced `MockKSettings.disallowMockingRestrictedClasses` to enforce strict mocking prevention.
- Setting can be configured via `settings.properties`:
  ```properties
  disallowMockingRestrictedClasses=false
  ```
- If set to true, attempting to mock a restricted class will throw an IllegalArgumentException.

### **3️⃣ Configurable Restricted Class List**

- Predefined restricted classes include:
  - System
  - Java Collections (List, Set, Map, Queue, etc.)
  - I/O classes (File, Path)
  - Value objects (default list can be extended by users)
  - Users can dynamically add/remove restricted classes.

### **4️⃣ Integration with mockk<T>()**
- `mockk<T>()` now checks for restricted classes before creating a mock.
- Calls RestrictedMockClasses.handleRestrictedMocking() to log warnings or throw exceptions based on settings.

### **5️⃣ Unit Tests** 
- Added tests to verify:
  - Warnings are logged when attempting to mock restricted classes.
  - Exceptions are thrown when disallowMockingRestrictedClasses = true.
  - User-defined restricted classes are correctly handled.


modules/mockk/src/commonMain/kotlin/io/mockk/impl/restrict/RestrictedMockClasses.kt
```diff
@@ -0,0 +1,49 @@
...
+            }
+        }
+    }
+
+    fun isRestricted(clazz: Class<*>): Boolean {
+        return defaultRestrictedTypes.any { it.isAssignableFrom(clazz) } ||
+                userDefinedRestrictedTypes.any { it.isAssignableFrom(clazz) }
+    }
+
+    fun addRestrictedType(clazz: Class<*>) {
```
Raibaz, line 38: What is the use case for changing the restricted types in code?

I'm worried that if we allow adding or removing a restricted type in code, tests will no longer be atomic.

Think of this case:

Test A mocks class Foo
Test B adds class Foo to the restricted types

Now the order of execution matters, because executing A before B passes and executing B before A doesn't, and yet tests should be independent.

How about making the user defined restricted types a configuration property?

Raibaz (changes requested):

modules/mockk/src/commonMain/kotlin/io/mockk/impl/restrict/RestrictedMockClasses.kt
devtaebong: I understand your concern about ensuring test independence and preventing runtime state modifications.  
I agree that managing user-defined restricted types through a configuration property would be a better approach.  

However, I wasn't able to locate a `settings.properties` file in the project.  
Could you clarify whether this file already exists but is currently inaccessible,  or if I'm simply missing it?  

If it's not available, would it make sense to create a new configuration properties file  specifically for managing the restricted types?  

Thanks in advance for your guidance! 🚀

devtaebong:
Hi @Raibaz ,

To guarantee test independence, I implemented MockingRestrictedExtension by extending BeforeEachCallback. This ensures that each test runs with a clean state.

Additionally, for better usability, I modified the API to allow users to:

- Register custom restricted classes via annotation (`@MockkRestricted`)
- Configure whether to throw an exception when mocking restricted classes directly through annotations.
Would love your feedback on this approach. Let me know if you see any improvements!

Thanks

Raibaz:
Thanks for looking into this!

I have a few concerns about this approach, though:
- IIUC, for mocking restrictions to work developers need to extend each test with `@MockingRestrictedExtension`, so if a developer tries to mock, say, `System` in a test that doesn't have the extension it will work without the restriction
- I'm not convinced about registering restricted classes via annotation: first of all, MockK is usually imported in projects with a `test` scope, so it is usually not visible to business code classes, and besides, developers may want to restrict mocking classes that belong to libraries, or that in general that are not their own; it wouldn't be possible this way

How about adding a configuration property in `MockKSettings` with the list of fully qualified names of user-defined restricted classes?

That would allow you to add, in your `mockk.properties` file, something like:

```
restrictedClasses=com.foo.Bar,com.foo.Baz
```

I know it's more error-prone than annotating restricted classes, but I think it gives the right amount of flexibility.

WDYT?

devtaebong:
> Thanks for looking into this!
> 
> I have a few concerns about this approach, though:
> 
> * IIUC, for mocking restrictions to work developers need to extend each test with `@MockingRestrictedExtension`, so if a developer tries to mock, say, `System` in a test that doesn't have the extension it will work without the restriction
> * I'm not convinced about registering restricted classes via annotation: first of all, MockK is usually imported in projects with a `test` scope, so it is usually not visible to business code classes, and besides, developers may want to restrict mocking classes that belong to libraries, or that in general that are not their own; it wouldn't be possible this way
> 
> How about adding a configuration property in `MockKSettings` with the list of fully qualified names of user-defined restricted classes?
> 
> That would allow you to add, in your `mockk.properties` file, something like:
> 
> ```
> restrictedClasses=com.foo.Bar,com.foo.Baz
> ```
> 
> I know it's more error-prone than annotating restricted classes, but I think it gives the right amount of flexibility.
> 
> WDYT?

Raibaz:
Oh, I see, that makes sense, thanks for clarifying!

So, I think that if a user restricts mocking a class in a MockKSetting, by default trying to mock it should throw an exception, or log a warning, anywhere in the application, without having to explicitly add the @MockingRestrictedExtension.

If anything, I'd like it to work the other way around: if class `Foo` is restricted from mocking it means you shouldn't try to mock it anywhere, but if you really really need to, you can override the restriction with a test extension, something like @AllowMockingRestrictedClasses.

What do you think about this approach?

Raibaz:
(also, thanks for the followup!)

devtaebong:
That makes a lot of sense! I see the value in making the restriction the default behavior while allowing an explicit override when absolutely necessary. Instead of relying solely on annotations, I’ll explore the possibility of managing restricted classes globally through a configuration file (e.g., mockk.properties). This would provide more flexibility while ensuring consistency across all tests.

I'll look into how this can be implemented effectively and update you with my findings. 
Thanks for the suggestion!

modules/mockk/src/commonMain/kotlin/io/mockk/MockK.kt
```diff
@@ -38,9 +39,10 @@ inline fun <reified T : Any> mockk(
     relaxed: Boolean = false,
     vararg moreInterfaces: KClass<*>,
     relaxUnitFun: Boolean = false,
-    block: T.() -> Unit = {}
+    mockValidator: MockkValidator = MockkValidator(RestrictMockkConfiguration()),
+    block: T.() -> Unit = {},
 ): T = MockK.useImpl {
-    RestrictedMockClasses.handleRestrictedMocking(T::class.java)
+    mockValidator.validateMockableClass(T::class)
```
devtaebong: I have added `mockValidator: MockkValidator = MockkValidator(RestrictMockkConfiguration())` as a parameter to the `mockk()` function.

`RestrictMockkConfiguration()` is designed to load the `mockk.properties` file from the user's classpath (`src/resources`) using a default PropertiesLoader parameter. Additionally, the implementation allows injecting a custom PropertiesLoader dynamically in tests.

Do you think this approach makes sense? Would you suggest any improvements?

modules/mockk/src/commonMain/kotlin/io/mockk/impl/restrict/propertiesloader/DefaultPropertiesLoader.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Properties
+
+class DefaultPropertiesLoader : PropertiesLoader {
+    override fun loadProperties(): Properties {
+        val properties = Properties()
+        val resourceStream = Thread.currentThread().contextClassLoader.getResourceAsStream("mockk.properties")
+        resourceStream?.use { properties.load(it) }
+        return properties
+    }
```
devtaebong, line 11: I haven't been able to properly test whether the file is correctly loaded from the user's classpath. Do you have any suggestions for a good approach?

devtaebong:
@Raibaz 
I've made modifications to load the restricted classes from the configuration file. I would greatly appreciate any additional feedback you may have.

Raibaz:
I like this!

I think we can merge this, but it needs two minor improvements:
- Can you please re-run `./gradlew :mockk-dsl:apiDump`?
- Can you please add documentation about how to use the mocking restriction in README.md?

Thanks a lot for looking into this! <3

devtaebong:
@Raibaz 

I've completed adding documentation on how to use mocking restrictions in `README.md`.

1. **Restricted Mocking Documentation:**
   - Added detailed explanations for **default restricted classes** and their behavior.
   - Clarified that **subclasses and implementations** are also restricted.

2. **Configuration Guide:**
   - Provided clear instructions on setting up `mockk.properties` for restricted classes.
   - Explained the behavior of `mockk.throwExceptionOnBadMock` with both default and custom settings.

3. **Custom Class Restriction Examples:**
   - Added usage examples for restricting **custom classes** (e.g., `com.foo.Bar`, `com.foo.Baz`).
   - Included both **exception-throwing** and **warning-logging** scenarios.

I'm really happy to contribute to a project I use frequently, and I appreciate all the valuable feedback you've provided. 😊  
Please let me know if there's anything else you'd like me to adjust or clarify!

Raibaz (approved):

README.md
```diff
@@ -1495,6 +1495,188 @@ So this is similar to the `returnsMany` semantics.
...
+
+- Log Example:
+  - List should not be mocked! Consider refactoring your test.
+
+#### When `mockk.throwExceptionOnBadMock=true`
+```kotlin
+@Test
+fun `when throwExceptionOnBadMock is true should throw MockKException for collections`() {
+    assertThrows<MockKException> {
+        mockk<List<String>>()  // Throws MockException
```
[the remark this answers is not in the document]
devtaebong: I’ve updated the comment to `MockException -> MockKException` as suggested. 😭

Raibaz:
Thanks a lot for the contribution and for the continued follow ups!

devtaebong:
Glad to hear you’re making good use of the feature!

mockk.properties is designed to be defined explicitly within each module. So in multi-module projects, you’ll need to place the file under each module’s test resources if you want module-specific configurations.

That said, it might be possible to set common defaults across modules using JVM system properties (e.g., via -D options). This could be a more scalable approach for large projects, but I’m not 100% sure if all settings from mockk.properties are respected via system properties — it would need some verification.

nishatoma:
@devtaebong Thank you so much for this work! I've tested it as well on multiple modules, and it looks like you can actually centralize the mockk properties file!

Some things I've observed while applying this change on a projects with big files: Lots of tests are failing now since naturally, there are some tests that mock a collection for example. 

Would you consider to introduce an option to gradually increase the restricted classes as opposed to starting with a full list?


For example, the idea is that for big orgs, we would start with no restrictions, and then over time, we can start adding collection, Map, etc. 

The current state is that there are a lot of tests that will fail, so this feature would need to be all/ or nothing if you have failing tests.

Another nice thing, is it would be really helpful and good QoL change to be able to isolate this feature for specific tests i.e Robolectric vs. pure JVM tests!

Let me know what you think, thank you again, really great work!

MarekMacko:
Could you share, @nishatoma, how you centralized the `mockk.properties` file?

nishatoma:
> Could you share, @nishatoma, how you centralized the `mockk.properties` file?

Sure! You need to have a module, call it "testing-module" for example. Place the mockk properties file there (within the src/main).

Then in your other modules, you could have them depend on your "testing-module" with a test implementation like so:

testImplementation project("testing-module")

All your modules that depend on that testing module should inherit that properties file by default.

Let me know if that helps!

MarekMacko:
> > Could you share, @nishatoma, how you centralized the `mockk.properties` file?
> 
> Sure! You need to have a module, call it "testing-module" for example. Place the mockk properties file there (within the src/main).
> 
> Then in your other modules, you could have them depend on your "testing-module" with a test implementation like so:
> 
> testImplementation project("testing-module")
> 
> All your modules that depend on that testing module should inherit that properties file by default.
> 
> Let me know if that helps!

The key point was to place it under `src/main`, not in `src/test`. Now it works like a charm! 
Thanks, @nishatoma 💪

Issue #1291, opened 2024-08-29
Allow suppressing "Failed to set backing field" OR fail test


cgm-aw:
### Introduction
Hi everyone,

I encountered a behavior of mockk, where, if the backing field has a different type than the getter, a warning is printed to the logs but the test continues and passes. 
This seems weird to me - I would either expect the test to fail or no warning message.
My suggestion:
Add a parameter to the mockk() function to be able to suppress the warning, and if suppression is not active, the test will fail (or the other way around).

I'm aware of this: https://mockk.io/#property-backing-fields
With the solution presented there, everything works fine. However, if you do not use this mechanism, you will just get confusing error messages and probably an inconsistent behavior.

### Prerequisites

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

Either a broken test or no log message

### Current Behavior

A warning message is printed, including a stack trace, but the test passes

#### Steps to Reproduce

Create this class (this is a Java class!)

```
import java.util.Optional;

public class ContainerClass {
    private Optional<String> someField;
    public String getSomeField() {
        return someField.orElse("empty");
    }
}
```

Create this Kotlin test:
```
    @kotlin.test.Test
    fun reproduceFailedToSetBackingField() {
        val testObject = mockk<ContainerClass>()

        every { testObject.someField } returns "mockValue"

    }
```

You will see this log entry:
```
2024-08-29 07:37:00,610 [main] WARN  i.m.i.r.states.StubbingAwaitingAnswerState - Failed to set backing field (skipping)
java.lang.IllegalArgumentException: Can not set java.util.Optional field test.ContainerClass.someField to java.lang.String
[...]
```

Using _propertyType_ yields the correct behavior:
```every { testObject.someField } propertyType Optional::class answers { "mockValue" }```

#### Context

* MockK version: 1.13.12
* OS: MacOS
* Kotlin version:2.0.20
* JDK version: 21
* JUnit version: 5.10.1
* Type of test: unit test

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
2024-08-29 07:37:00,610 [main] WARN  i.m.i.r.states.StubbingAwaitingAnswerState - Failed to set backing field (skipping)
java.lang.IllegalArgumentException: Can not set java.util.Optional field test.ContainerClass.someField to java.lang.String
[...]

PR #1349, opened 2025-02-17
Added new property "failOnSetBackingFieldException" to fail test if a backing field could not be set

cgm-aw:
Hi,

as mentioned in https://github.com/mockk/mockk/issues/1291, if setting a backing field fails, it just prints a warning but the test does not fail. I introduced a new parameter to alter this behavior - when the new property is set to true, the test fails.
I think this improves handling a lot, as clearly the developer wanted to mock something which failed.
The parameters are a bit confusing to me, as there is the io/mockk/settings.properties and the mockk.properties file. I used settings.properties because it is available in the code I added the fix to. The property "throwExceptionOnBadMock" sounds like it should do what I implemented, but I did not want to break existing codebases by re-using this property.
The unit tests are not very nice, the real problem occurs if you mix Java and Kotlin code (see the original issue) and Java files are not available on the classpath in the current setup of MockK.

Please check it out, I would be happy if this change made it :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1309,10 +1309,14 @@ relaxUnitFun=true|false
 recordPrivateCalls=true|false
 stackTracesOnVerify=true|false
 stackTracesAlignment=left|center
+failOnSetBackingFieldException=true|false
 ```
 
-`stackTracesAlignment` determines whether to align the stack traces to the center (default),
+* `stackTracesAlignment` determines whether to align the stack traces to the center (default),
  or to the left (more consistent with usual JVM stackTraces).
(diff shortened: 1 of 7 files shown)


Raibaz:
Thanks for looking into this!

Overall the PR looks good to me.

Could you also add the documentation about the new configuration flag, please?

Thanks!

cgm-aw:
Hi,

sure, please check the change: https://github.com/mockk/mockk/commit/280047bf82b65cbb7fb1010a4320df37a4f2efea

Best regards

PR #1366, opened 2025-03-27
Do not generate visibility bridge for package-private default methods (fixes #1118)

### LINKED ISSUE #1118: Mocking jOOQ 3.15+ fails with "attempted to add a method"
This issue is the continuation of the analysis started and documented at https://stackoverflow.com/q/76665966/131929. Also, this is very closely related to #255 and #477 which both became stale.

I tried to debug further but the exception stems from native JVM code. Hence, I suspect it requires someone intimately familiar with this project to look into it. Allow me to tag the top contributors @oleksiyp, @aSemy and @Raibaz.

### Prerequisites

### Expected Behavior

Mocking the jOOQ 3.15+ `TableImpl` class should work.

### Current Behavior

It fails with

> java.lang.UnsupportedOperationException: class redefinition failed: attempted to add a method

### Failure Information (for bugs)

Mocking the same class with jOOQ 3.14 works fine. The 3.15+ versions are different in that they also implement interfaces which seems to trip up MockK. With the earlier jOOQ versions MockK prints this trace log

> Taking instance of class org.jooq.impl.TableImpl itself because it is not abstract and no additional interfaces specified.

It fails both on Java 11 and Java 17.

#### Steps to Reproduce

Check out the reproducer at https://github.com/marcelstoer/jooq-mockk-test and run `mvn test`.

sgerke-1L:
Minimal reproducer for issue #1118 with fix.

The root problem is that the JVM doesn't generate a visibility bridge for package-private default methods, but byte-buddy does. This was discussed in https://github.com/raphw/byte-buddy/issues/1795. 

It might be that we can use `decorate` instead of `redefine` to side-step the problem, but I'm not experienced enough with mockk to judge if this can work.

The previous approach of not generating the bridge methods by deselecting them with a custom visibility bridge strategy can be found in the commit history and might be an alternative if `decorate` is not viable.


marcelstoer:
@sgerke-1L thanks for your analysis!

By comparing the Javadocs for the two methods I concluded that the existing `redefine()` is what we likely want.

| Decorate | Redefine |
| --- | --- |
| Decorates a type with `net.bytebuddy.asm.AsmVisitorWrapper` and allows adding attributes and annotations. A decoration does not allow for any standard transformations but can be used as a performance optimization compared to a redefinition, especially when implementing a Java agent that only applies ASM-based code changes.<br>Important: Only use this mode to improve performance in a narrowly defined transformation. Using other features as those mentioned might result in an unexpected outcome of the transformation or error. Using decoration also requires the configuration of an `Implementation.Context.Factory` that does not attempt any type transformation. | Redefines the given type where any intercepted method that is declared by the redefined type is fully replaced by the new implementation.<br>Note: When a user redefines a class with the purpose of reloading this class using a `net.bytebuddy.dynamic.loading.ClassReloadingStrategy`, it is important that no fields or methods are added to the redefined class. Note that some `Implementation`s implicitly add fields or methods. Finally, Byte Buddy might be forced to add a method if a redefined class already defines a class initializer. This can be disabled by setting `with(Implementation.Context.Factory)` to use a `Implementation.Context.Disabled.Factory` where the class initializer is retained as is. |

@Raibaz can we ask you to approve the workflow run to get full test coverage, thanks.

Raibaz:
Sure, and thanks a lot to both of you for the in-depth analysis!

marcelstoer:
After revisiting raphw/byte-buddy#1795 I changed my opinion. True, the Byte Buddy Javadoc does warn against the use of `decorate()` except for a limited set of use cases. Ours, while not for performance reasons, seems to fall into exactly this category as we only use the decorator for the visitor pattern.

I'd be happy to see this land and released soon.

sgerke-1L:
Thanks for your support @marcelstoer. I still can't judge the full consequences of using `decorate` over `redefine` (as I don't know all details of how mockk is interacting with the decorated/redefined class), but I also think it might be worth a try. If there are examples in the wild that this change breaks, then there will be bug reports and new test-cases to be added.

sgerke-1L:
Btw, is it possible to run the android tests as well or would that have to be done locally?

Raibaz:
The instrumented tests for api level 30 have been run, I see that as the first check in the list.

Issue #821, opened 2022-05-16
ConfirmVerified not isolated per tests


dekar91:
Hi!
I am not sure if it is desired behavior, but it started failing since `1,2.4`. Now, `confirmVerified` fails not only if there are no verified calls in the current test but also in the previous one.

### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior

Expected, `confirmVerified` checks only mock of the current scope.

### Current Behavior

`confirmVerified` fails if there were unchecked calls in previous tests

#### Context

* MockK version: 1.2.14
* OS: Ubuntu 22 LTS
* Kotlin version: 1.6.21
* JDK version: 17
* JUnit version: 5 (koTests)
* Type of test: unit test

#### Failure Logs

<details>
<summary>Stack trace</summary>

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------

> Task :processResources NO-SOURCE
> Task :processTestResources NO-SOURCE
> Task :compileKotlin
w: /home/dekar/projects/kotlin-flow/src/main/kotlin/Main.kt: (7, 10): Parameter 'args' is never used
> Task :compileJava NO-SOURCE
> Task :classes UP-TO-DATE
> Task :compileTestKotlin
> Task :compileTestJava NO-SOURCE
> Task :testClasses UP-TO-DATE
> Task :test
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended

Verification acknowledgment failed

Verified call count: 0
Recorded call count: 1


Not verified calls:
1) SupplierInterface(#1).supply()

Stack traces:
1)                                                       io.mockk.impl.InternalPlatform.captureStackTrace                      (InternalPlatform.kt:125)                    
                                                           io.mockk.impl.stub.MockKStub.handleInvocation                       (MockKStub.kt:250)                           
                         io.mockk.impl.instantiation.JvmMockFactoryHelper$mockHandler$1.invocation                             (JvmMockFactoryHelper.kt:23)                 
                                                  io.mockk.proxy.jvm.advice.Interceptor.call                                   (Interceptor.kt:21)                          
                                                   io.mockk.proxy.jvm.advice.BaseAdvice.handle                                 (BaseAdvice.kt:42)                           
[...]

PR #1367, opened 2025-03-28
fix: ensure confirmVerified is isolated per test

rsvinicius:
## Pull Request Description  

This PR implements the approach discussed in **issue #821**—**resetting recorded calls between test cases** to ensure test isolation without affecting shared mocks.  

## Changes Made  
- Updated `internalConfirmVerified` to **reset recorded calls** while keeping shared mocks intact.  
- Used `ClearOptions(recordedCalls = true, verificationMarks = true, …)` to ensure test isolation **without affecting persistent mocks**.  
- Performed some **refactoring** to improve code clarity—please let me know if these changes are acceptable.  

### 1. Test Failure in `clearMarks()`  
This test fails under the new behavior because it expects **manually cleared verification marks** to be considered.  
The new approach **prevents clearing unless explicitly requested**, causing this assertion to fail:  

```kotlin
@Test
fun clearMarks() {
    every { mock.op(1) } returns 1
    every { mock.op(2) } returns 2
    every { mock.op(3) } returns 3
    every { mock.op(4) } returns 4

    assertEquals(1, mock.op(1))
    assertEquals(2, mock.op(2))
    assertEquals(3, mock.op(3))
    assertEquals(4, mock.op(4))

    verify {
        mock.op(1)
        mock.op(2)
        mock.op(3)
    }

    assertFails {
        confirmVerified(mock)
    }

    verify {
        mock.op(4)
    }

    confirmVerified(mock)

    clearMocks(
        mock,
        answers = false,
        recordedCalls = false,
        childMocks = false,
        verificationMarks = true,
        exclusionRules = false
    )

    assertFails {
        confirmVerified(mock)
    }
}
```

Should this test be updated to **align with the new behavior**, or do I need to handle **verification marks differently**?  

### 2. Refactoring Changes  
I made some **refactoring** in the `internalConfirmVerified` method—are these changes acceptable?

### DIFF
--- a/modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
+++ b/modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
@@ -270,13 +270,27 @@ object MockKDsl {
      * Checks if all recorded calls were verified.
      */
     fun internalConfirmVerified(mocks: Array<out Any>) {
+        val verifier = MockKGateway.implementation().verificationAcknowledger
+
         if (mocks.isEmpty()) {
-            MockKGateway.implementation().verificationAcknowledger.acknowledgeVerified()
+            verifier.acknowledgeVerified()
+        } else {
+            mocks.forEach { verifier.acknowledgeVerified(it) }
         }
 
-        for (mock in mocks) {
-            MockKGateway.implementation().verificationAcknowledger.acknowledgeVerified(mock)
-        }
+        resetVerificationState()
+    }
+
+    private fun resetVerificationState() {
+        val options = ClearOptions(
+            answers = false,
+            recordedCalls = true,
+            childMocks = false,
+            verificationMarks = true,
+            exclusionRules = false
+        )
+
+        MockKGateway.implementation().clearer.clearAll(options, currentThreadOnly = true)
     }
 
     /**
--- a/modules/mockk/src/jvmTest/kotlin/io/mockk/it/VerificationAcknowledgeTest.kt
+++ b/modules/mockk/src/jvmTest/kotlin/io/mockk/it/VerificationAcknowledgeTest.kt
@@ -632,7 +632,6 @@ class VerificationAcknowledgeTest {
 
     @Test
     fun confirmVerifiedAll() {
-        clearAllMocks()
         doCalls1()
 
         verify {
@@ -646,7 +645,6 @@ class VerificationAcknowledgeTest {
 
     @Test
     fun confirmVerifiedAllInverse() {
-        clearAllMocks()
         doCalls1()
 
         verify {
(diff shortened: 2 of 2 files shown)


Raibaz:
Thanks for looking into this!

Changes look good to me.

I think the failing test can be removed altogether: manually clearing verification marks inside a test is a behavior seems like a very weird behavior to me and it's ok to me not to account for it in our tests.

rsvinicius:
@Raibaz Thank you for reviewing!  

I've removed the `clearMarks` test as suggested. Let me know if there's anything else I should adjust.  

Appreciate your time!

davidconnard:
Hey @rsvinicius 

I've tracked down this change causing failures in some of our projects existing tests.  We previously had tests that called `confirmVerified(mock1)` and _then_ called `coVerify { mock2.foo() }`.  It seems that, after the 1.14.0 update, that this pattern is no longer valid test code.  It seems that `confirmVerified(...)` must be called last, and with _all_ mocks, ... ie: `confirmVerified(mock1)` followed by confirmVerified(mock2)` (or any other verify statement) will never work.

Is this expected?  If so - this should probably be called out specifically as part of the changelogs - it took me a while to track down.  If not, can we do better here, and meet the intent of this change, but yet allow the current test to continue to work with any verification order.

rsvinicius:
> Hey @rsvinicius
> 
> I've tracked down this change causing failures in some of our projects existing tests. We previously had tests that called `confirmVerified(mock1)` and _then_ called `verify { mock2.foo() }`. It seems that, after the `1.14.0` update, that this pattern is no longer valid test code. It seems that `confirmVerified(...)` must be called last, and with _all_ mocks, ... ie: `confirmVerified(mock1)` followed by `confirmVerified(mock2)` (or any other verify statement) will never work. Re-ordering our tests so that `confirmVerified(..)` is called last seems to make them work again under `1.14.0`.
> 
> Is this expected? If so - this should probably be called out specifically as part of the changelogs - it took me a while to track down. If not, can we do better here, and meet the intent of this change, but yet allow the current test to continue to work with any verification order?

Hey @davidconnard , thanks for the detailed feedback!

You're right — the behavior introduced in 1.14.0 has led to some unintended side effects, particularly when calling confirmVerified(mock1) before verifying other mocks. This can break tests that previously worked in 1.13.x.

There’s an open issue tracking this regression: [#1379](https://github.com/mockk/mockk/issues/1379), which was raised by another community member after encountering similar issues.

I’ve been looking into it and trying to make confirmVerified(...) only reset invocations for the mocks explicitly passed in. But I'm running into some odd behavior when using confirmVerified() with no arguments — it doesn't always behave consistently, especially when mixed with other verifications (across tests).

Thank you again for helping surface and confirm this problem.

Issue #1375, opened 2025-04-10
Version 1.14.0 introduces `class redefinition failed: invalid class` in Android local tests


bddckr:
On MockK version 1.13.17, this code in the `test` folder of an Android library project works:
```kotlin
package mypackage

import android.net.Uri
import io.mockk.every
import io.mockk.mockk
import io.mockk.mockkStatic
import io.mockk.unmockkStatic
import org.junit.jupiter.api.Test

class Test {
    @Test
    fun test() {
        try {
            mockkStatic(Uri::class)
            every { Uri.parse(any()) } answers { mockk() }
        } finally {
            unmockkStatic(Uri::class)
        }
    }
}
```

But in MockK 1.14.0 it fails:
```
class redefinition failed: invalid class
java.lang.InternalError: class redefinition failed: invalid class
	at java.instrument/sun.instrument.InstrumentationImpl.retransformClasses0(Native Method)
	at java.instrument/sun.instrument.InstrumentationImpl.retransformClasses(Unknown Source)
	at io.mockk.proxy.jvm.transformation.JvmInlineInstrumentation.retransform(JvmInlineInstrumentation.kt:28)
	at io.mockk.proxy.common.transformation.RetransformInlineInstrumentation$doCancel$1.invoke(RetransformInlineInstrumentation.kt:38)
	at io.mockk.proxy.common.transformation.RetransformInlineInstrumentation$doCancel$1.invoke(RetransformInlineInstrumentation.kt:32)
	at io.mockk.proxy.common.transformation.ClassTransformationSpecMap.applyTransformation(ClassTransformationSpecMap.kt:41)
	at io.mockk.proxy.common.transformation.RetransformInlineInstrumentation.doCancel(RetransformInlineInstrumentation.kt:32)
	at io.mockk.proxy.common.transformation.RetransformInlineInstrumentation.access$doCancel(RetransformInlineInstrumentation.kt:6)
	at io.mockk.proxy.common.transformation.RetransformInlineInstrumentation$execute$1$1.invoke(RetransformInlineInstrumentation.kt:17)
	at io.mockk.proxy.common.transformation.RetransformInlineInstrumentation$execute$1$1.invoke(RetransformInlineInstrumentation.kt:17)
	at io.mockk.proxy.common.CancelableResult.cancel(CancelableResult.kt:22)
	at io.mockk.proxy.common.CancelableResult$alsoOnCancel$1.invoke(CancelableResult.kt:30)
[...]

PR #1376, opened 2025-04-11
Use redefine if mockkStatic is used (fix candidate for #1375)

sgerke-1L:
Based on the examples provided in #1375, this MR contains a reproducer for jacoco coverage collection that fails due to and after #1366.

It also contains a proposed fix: The fix is to revert to using redefine if static mockking is required. This is an obvious proposal that satisfies all our test-cases, but it does not answer the bigger question whether decorate is the right approach anyway (or whether we should use redefine in all cases, and omit visibility bridges for default methods).

So basically the two approaches boil down to changing the instrumentation code to
```kotlin
if (spec.shouldDoStaticIntercept) {
    redefine(classBeingRedefined, of(classBeingRedefined.name, classfileBuffer))
} else {
    decorate(classBeingRedefined, of(classBeingRedefined.name, classfileBuffer))
}
```
or
```kotlin
.with(VisibilityBridgeStrategy { not(isDefaultMethod()).matches(it) })
.redefine(classBeingRedefined, of(classBeingRedefined.name, classfileBuffer))
```

Both fix the reproducer I've extracted.

TODOs:

- [ ] Agree on approach
- [ ] Review jacoco integration

### DIFF
--- a/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
+++ b/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
@@ -42,8 +42,6 @@ object Deps {
         const val junit4 = "junit:junit:${Versions.junit4}"
         const val junitJupiter = "org.junit.jupiter:junit-jupiter:${Versions.junitJupiter}"
 
-        const val apacheCommons = "org.apache.commons:commons-lang3:${Versions.apacheCommons}"
-
         const val kotlinCoroutinesBom = "org.jetbrains.kotlinx:kotlinx-coroutines-bom:${Versions.coroutines}"
         const val kotlinCoroutinesCore = "org.jetbrains.kotlinx:kotlinx-coroutines-core"
         const val kotlinCoroutinesTest = "org.jetbrains.kotlinx:kotlinx-coroutines-test"
--- a/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
+++ b/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
@@ -18,7 +18,6 @@ object Deps {
         const val logback = "1.4.5"
         const val junitJupiter = "5.8.2"
         const val junit4 = "4.13.2"
-        const val apacheCommons = "3.17.0"
 
         const val byteBuddy = "1.14.17"
         const val objenesis = "3.3"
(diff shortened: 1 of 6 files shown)


Raibaz:
I agree with @marcelstoer.

Wondering if we're able to get @raphw's opinion on this.

sgerke-1L:
Thanks for your input @marcelstoer and @Raibaz. Another argument for giving the solution with the `VisibilityBridgeStrategy` a try is that that would keep using `redefine`, and that minimizes the change to what we had before.

I've updated this MR according to your input in the meantime. 

Let me know if you have further comments. I am also looking for feedback regarding the integration of jacoco. I think we might currently be running jacoco and kover both for the client-tests. Let me know if you have suggestions here.

raphw:
If you only ever want to change code but never add fields or methods, why not always decorate? No need for a condition and some general performance improvement.

As for visibility bridges: that Byte Buddy fixes up the oversight of javac is more of an accident as Byte Buddy just tries to stick to the spec by default. I'd disable in general for your case, not just for default methods. If they are already there, Byte Buddy won't touch them.

sgerke-1L:
First of all, thanks for your helpful and sound advice in this matter, @raphw 

The answer to both of your questions is basically Hyrum's Law.

> If you only ever want to change code but never add fields or methods, why not always decorate? No need for a condition and some general performance improvement.
Apparently, jacoco is depending on some implementation detail of `redefine` (for static mocks), and breaks when we use `decorate`. 

> As for visibility bridges: that Byte Buddy fixes up the oversight of javac is more of an accident as Byte Buddy just tries to stick to the spec by default. I'd disable in general for your case, not just for default methods. If they are already there, Byte Buddy won't touch them.
Again, Hyrum's law. The change as I'm proposing it here is the smallest I know to fix the original issue #1118. By disabling all bridge methods, we'd introduce another change from the state from before, so let's just not do this.

I'd suggest trying the change as proposed here next and see if we learn about more issues. Since we want to provide a non-major release, we should minimize breakage. We got reports that the previous fix (although probably technically sound) introduced regressions in projects that also apply instrumentation, because there the surface for interaction may include properties of how the instrumentation is done in detail.

Looking forward to hearing your thoughts!

raphw:
If you retransform previously loaded classes, you will not be able to add fields or methods. Decoration implies this limitation, so I would just go for it as this only reflects a well-tested performance improvement.

marcelstoer:
Thanks @Raibaz, looking forward to the new release.

PR #1378, opened 2025-04-11
Change mockk configuration paths to test sourceSet

TWiStErRob:
Do not recommend users to change (and ship!) test configuration in their production builds.


Raibaz:
Good catch, thanks for this!

Issue #1379, opened 2025-04-15
Version 1.14.0: confirmVerified clears invocation of other mocks


MalteJoe:
With the upgrade to 1.14.0 some of our tests broke, but they probably shouldn't. This is a reproducible sample:

```kotlin
import io.mockk.*
import io.mockk.impl.annotations.MockK
import io.mockk.junit5.MockKExtension
import org.junit.jupiter.api.Test
import org.junit.jupiter.api.extension.ExtendWith

@ExtendWith(MockKExtension::class)
class ReproductionTest {
    interface SomeService {
        fun doSomething()
        fun doSomethingElse()
    }

    @Test
    fun `this worked with 1-13-17`(@MockK someMockA: SomeService, @MockK someMockB: SomeService) {
        every { someMockA.doSomething() } just runs
        every { someMockB.doSomethingElse() } just runs

        someMockA.doSomething()
        someMockB.doSomethingElse()

        verify { someMockA.doSomething() }
        confirmVerified(someMockA)
        // fails: Verification failed: call 1 of 1: SomeService(#6).doSomethingElse()) was not called
        verify { someMockB.doSomethingElse() }
        confirmVerified(someMockB)
    }

    @Test
    fun `this works with 1-14-0`(@MockK someMockA: SomeService, @MockK someMockB: SomeService) {
        every { someMockA.doSomething() } just runs
        every { someMockB.doSomethingElse() } just runs

        someMockA.doSomething()
        someMockB.doSomethingElse()

        // Also works with separate verify checks like above
        verify {
            someMockA.doSomething()
            someMockB.doSomethingElse()
        }

        // works too: confirmVerified(someMockA, someMockB)
        confirmVerified(someMockA)
        confirmVerified(someMockB)
    }
}
```

Starting with `1.14.0`, the first test fails on the line `verify { someMockB.doSomethingElse() }`: (Both tests run fine with `1.13.17`)

```
java.lang.AssertionError: Verification failed: call 1 of 1: SomeService(#6).doSomethingElse()) was not called

	at io.mockk.impl.recording.states.VerifyingState.failIfNotPassed(VerifyingState.kt:61)
	at io.mockk.impl.recording.states.VerifyingState.recordingDone(VerifyingState.kt:40)
	at io.mockk.impl.recording.CommonCallRecorder.done(CommonCallRecorder.kt:47)
	at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:63)
[...]

PR #1381, opened 2025-04-22
fix: prevent confirmVerified(vararg mocks) from clearing unrelated mocks

rsvinicius:
## 🚧 Draft Pull Request – Partial Fix for Issue [#1379](https://github.com/mockk/mockk/issues/1379)

This PR builds upon the changes introduced in [#821](https://github.com/mockk/mockk/issues/821), aiming to improve test isolation while addressing the regression reported in [#1379](https://github.com/mockk/mockk/issues/1379).

### 🛠 What's Being Done

- Refactored `internalConfirmVerified` to **reset only the specified mocks** when parameters are passed.
- Updated `resetVerificationState` to **selectively clear** verification state for specific mocks.
- Ensured cleanup with `try-finally`, making the verification process more robust.
- Added a test case `confirmVerifiedSpecificMock` to verify that `confirmVerified(mock)` behavior is correct.

### ⚠️ Current Status

This PR is still **in progress**. While the targeted clearing works in isolated tests, the test suite fails when running `confirmVerifiedAll`.

This issue appears similar to the cross-test interference previously addressed in issue [#821](https://github.com/mockk/mockk/issues/821), but now occurs when `confirmVerified` is used without parameters after using it with specific mocks in other tests.

👉 Notably, if we add a `clearAllMocks()` call before the `confirmVerifiedAll` test, it passes successfully — just like it did before PR [#1367](https://github.com/mockk/mockk/pull/1367).

### 🔍 Next Steps
Investigate and resolve the failure of `confirmVerifiedAll` in the full test suite  

@Raibaz  do you have any idea why this might be happening? It seems related to state retention when mixing `confirmVerified(mock)` and `confirmVerified()` in different tests.

### DIFF
--- a/modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
+++ b/modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
@@ -272,16 +272,19 @@ object MockKDsl {
     fun internalConfirmVerified(mocks: Array<out Any>) {
         val verifier = MockKGateway.implementation().verificationAcknowledger
 
-        if (mocks.isEmpty()) {
-            verifier.acknowledgeVerified()
-        } else {
-            mocks.forEach { verifier.acknowledgeVerified(it) }
+        try {
+            if (mocks.isEmpty()) {
+                verifier.acknowledgeVerified()
+            } else {
+                mocks.forEach { verifier.acknowledgeVerified(it) }
+            }
+        } finally {
+            resetVerificationState(mocks)
         }
-
-        resetVerificationState()
     }
 
-    private fun resetVerificationState() {
+    private fun resetVerificationState(mocks: Array<out Any>) {
+        val implementation = MockKGateway.implementation()
         val options = ClearOptions(
             answers = false,
             recordedCalls = true,
@@ -290,7 +293,11 @@ object MockKDsl {
             exclusionRules = false
         )
 
-        MockKGateway.implementation().clearer.clearAll(options, currentThreadOnly = true)
+        if (mocks.isEmpty()) {
+            implementation.clearer.clearAll(options, false)
+        } else {
+            implementation.clearer.clear(mocks, options)
+        }
     }
 
     /**
--- a/modules/mockk/src/jvmTest/kotlin/io/mockk/it/VerificationAcknowledgeTest.kt
+++ b/modules/mockk/src/jvmTest/kotlin/io/mockk/it/VerificationAcknowledgeTest.kt
@@ -629,5 +629,23 @@ class VerificationAcknowledgeTest {
 
         confirmVerified()
     }
+
+    @Test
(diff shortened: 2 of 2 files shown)


modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
```diff
@@ -290,7 +293,11 @@ object MockKDsl {
             exclusionRules = false
         )
 
-        MockKGateway.implementation().clearer.clearAll(options, currentThreadOnly = true)
+        if (mocks.isEmpty()) {
+            implementation.clearer.clearAll(options, false)
```
Raibaz, line 297: Why was this changed from true to false?

Raibaz:
Wild guess: could it be that `verify` with more than one call, such as the one in the `confirmVerifiedAll` test, don't add the right values to the list of verified calls?

Can you check the values of `allCalls` and `verifiedCalls` [here](https://github.com/mockk/mockk/blob/23038a0f2519aa466eb2f8c187dc6593d52e4aa8/modules/mockk/src/commonMain/kotlin/io/mockk/impl/recording/CommonVerificationAcknowledger.kt#L42), when your test fails?

modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
rsvinicius: To be honest, I’m not entirely sure which value is more appropriate here. I noticed that clearAllMocks uses currentThreadOnly = false, so I followed that behavior for consistency. But if there's a specific reason it should be true, I’m happy to revisit this.

rsvinicius:
> Wild guess: could it be that `verify` with more than one call, such as the one in the `confirmVerifiedAll` test, don't add the right values to the list of verified calls?

I don’t think that’s the root cause, because when the confirmVerifiedAll test is run in isolation, it passes — so it seems that verify { ... } with multiple calls is working as expected in that context.

However, when the whole test suite is executed, the same test fails. That suggests some shared state leakage or interference between tests, similar to what we saw before the fix for #821.


> Can you check the values of `allCalls` and `verifiedCalls` [here](https://github.com/mockk/mockk/blob/23038a0f2519aa466eb2f8c187dc6593d52e4aa8/modules/mockk/src/commonMain/kotlin/io/mockk/impl/recording/CommonVerificationAcknowledger.kt#L42), when your test fails?

Yes — I added debug prints and here’s what I found during a failing run (screenshot attached below). So it looks like the verifiedCalls list gets cleared unexpectedly across tests. Interestingly, if I add a clearAllMocks() at the beginning of the confirmVerifiedAll test, it passes — just like it used to before PR #1367.

image

Raibaz:
That's interesting - I'd do more investigation to understand who's clearing the verifiedCalls list, when and why :)

Issue #1384, opened 2025-04-28
Add support for meta annotations


smilasek:
Hey,

In couple of our projects we were trying to use mockk in a following way:
```
@Retention(RUNTIME)
@Target(TYPE)
@CheckUnnecessaryStub
@ExtendWith(MockKExtension.class)
public @interface MockkTest {
}
```
Our goal is to be able to quickly annotate all of our tests by simply `@MockkTest`, which would configure some default behaviour like checking for unnecessary stubs and eventually more. This way we could end up with multiple annotations like `@MockWithStubsCheck`, @MockWithVerificationEnabled` etc

Unfortunately, after some testing, I have found out that `MockkExtension` is checking for the presence of an annotation on a class or eventually looking for a configured property. Snippet below:
```
    private val ExtensionContext.checkUnnecessaryStub: Boolean
        get() = testClass.checkUnnecessaryStub ||
                getConfigurationParameter(CHECK_UNNECESSARY_STUB_PROPERTY).map { it.toBoolean() }.orElse(false)

    private val Optional<out AnnotatedElement>.checkUnnecessaryStub
        get() = map { it.getAnnotation(CheckUnnecessaryStub::class.java) != null }
            .orElse(false)
```
So, in a nutshell, it does not have any support for meta annotations. Is there any big deal for you to add that? 

Currently, we are trying to mitigate it by adding the following snippet to `build.gradle`:
```
  tasks.withType(Test).configureEach {
    systemProperty "mockk.junit.extension.checkUnnecessaryStub", "true"
  }
```
but its not scalable and we would like to eventually have that annotation as a part of some internal lib so we could reuse it in various places.

Thanks for the feedback

Raibaz:
I have nothing against it: in fact, it seems like a good feature addition to me 👍 

I'd gladly review a PR about it and the implementation doesn't even seem to complex, so I'll mark this as a good first issue.

VitalyVPinchuk:
Hi, can I try to implement it?
@Raibaz

Raibaz:
Sure, go ahead!

shamim-emon:
Is this still available? asking this because eventhough this issue is open, I see #1393 addressing this

VitalyVPinchuk:
It's merged, I just forgot to indicate `fix #issue`

PR #1393, opened 2025-05-12
Add recursive detection of meta-annotations

VitalyVPinchuk:
Adds support for recursive meta-annotation detection, as described in #1384.


Raibaz:
Thanks for looking into this!

PR #1394, opened 2025-05-12
Update Android aritfacts to support 16kb page sizes

### LINKED ISSUE #1392: provide 16KB binaries for android
Using the new agp 8.10.0 there is a new lint warning introduced: Aligned16KB
used mockk version: 1.14.2

Native library dependency not 16 KB aligned The native library `arm64-v8a/libmockkjvmtiagent.so` (from `io.mockk:mockk-agent-android:1.14.2`) is not 16 KB aligned Android has traditionally used 4 KB memory page sizes. However, to support future devices that only work with 16 KB aligned libraries apps containing native libraries need to be built with 16 KB alignment. Apps with 4 KB aligned native libraries may not work correctly on devices requiring 16 KB alignment. To ensure compatibility and future-proof your app, it is strongly recommended that your native libraries are aligned to 16 KB boundaries. If your app uses any NDK libraries, directly or indirectly through an SDK, you should rebuild your app to meet this recommendation. Make sure all native libraries within your application, including those from dependencies, are built with 16 KB page alignment. This lint check looks at all native libraries that your app depends on. If any are found to be aligned to 4 KB instead of 16 KB, you will need to address this. When a library is flagged, first try to update to a newer version that supports 16 KB alignment. If an updated version is not available, contact the library vendor to ask about their plans for 16 KB support and request a compatible version. Updating your libraries proactively will help ensure your app works properly on a wider range of devices.

WhosNickDoglio:
Fixes #1392

https://developer.android.com/guide/practices/page-sizes#compile-16-kb-alignment

### DIFF
--- a/modules/mockk-agent-android/CMakeLists.txt
+++ b/modules/mockk-agent-android/CMakeLists.txt
@@ -31,3 +31,5 @@ add_library(mockkjvmtiagent
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -frtti -Wall -Werror -Wno-unused-parameter -Wno-shift-count-overflow -Wno-error=non-virtual-dtor -Wno-sign-compare -Wno-switch -Wno-missing-braces")
 
 target_link_libraries(mockkjvmtiagent slicer)
+
+target_link_options(mockkjvmtiagent PRIVATE "-Wl,-z,max-page-size=16384")


Raibaz:
@WhosNickDoglio thanks for looking into this! Is this ready for review?

Raibaz:
Thanks!

Issue #707, opened 2021-09-30
Message from Assertk assertions is not printed when using withArg block.


CamilYed:
### Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Expected Behavior
The message from AssertionFailedError (which is thrown by Assertk lib) should be printed when assertion is performed in **withArg** block.

### Current Behavior
The message from AssertionFailedError (which is thrown by Assertk lib) exception is not printed when the checking is performed in **withArg** block.

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.11.0
* OS: macOS Big Sur
* Kotlin version: 1.5
* JDK version: 11
* JUnit version: 5
* Type of test: unit test

#### Minimal reproducible code (the gist of this issue)

```kotlin
// -----------------------[ GRADLE DEFINITIONS ] -----------------------

project.ext.versions = [
  junit                : '5.7.2',
  assertk           : '0.24',
  mockk            : '1.11.0'
]

dependencies {
  testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: versions.junit
  testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: versions.junit
  testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params', version: versions.junit
  testImplementation group: 'com.willowtreeapps.assertk', name: 'assertk-jvm', version: versions.assertk
  testImplementation group: 'io.mockk', name: 'mockk', version: versions.mockk
}
// -----------------------[ YOUR CODE STARTS HERE ] -----------------------
package some.example.com

import io.mockk.every
import io.mockk.mockk
import org.apache.kafka.common.KafkaFuture
import kotlin.test.Test

class SomeExampleTest {

    val orderRepository: OrderRepository = mockk()
    val orderService: OrderService = OrderService(orderRepository)

    @Test
    fun test() {
       // given
       thereIsOrder(anOrder())

      // when
      orderService.pay(anOrderPayment())

      // then
      wasSaved(anOrder())
    }
    
    fun wasSaved(order: Order, times: Int = 1) {
      verify(exactly = times) {
         orderRepository.save(withArg{ equalsTo(order) })
      }
    }
    
    private fun MockKAssertScope.equalsTo(order: Order) {
[...]

PR #1395, opened 2025-05-17
Add logging on failure inside withArg

Djaler:
One more fix for #707

Previous fix (#776) was reverted (https://github.com/mockk/mockk/issues/707#issuecomment-1056006754), but this one should not repeat the same error.

I use the same idea as in mockito-kotlin `check`: https://github.com/mockito/mockito-kotlin/blob/main/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Verification.kt#L285

It is not ideal, as it may print assertion errors even on verify pass (in case on many invocations), but this is how mockit-kotlin `check` works too. And I think this is better than nothing


Djaler:
@Raibaz hello, can you look at this, please?

Djaler:
Can you release this, please?

Raibaz:
Done!

PR #1396, opened 2025-05-18
fix: enable confirmVerified to work with static mocks

### LINKED ISSUE #242: Bug: enable confirmVerified to work with static mocks
Enable to verify a static mock is not called anymore with the routine confirmVerified

`mockkStatic(Session::class)`
`every { Session.getToken() } returns "tokenSso"`
`...`
`confirmVerified(Session::class)`

happysubin:
fixes #242

When a static class is mocked using the mockStatic method, its javaClass is used as the key in the stubRepo. 
To make use of this, add logic to convert a given KClass to its javaClass and use it in the when clause.


Raibaz:
This looks good, thanks for looking into it!

PR #1418, opened 2025-07-25
make captured value reference volatile

mmimica:
I'm getting `Value not yet captured.` at random. It's after the mock is verified, so it has been called for sure.


Raibaz:
Hi, thanks for looking into this!

Would you mind adding test coverage?

mmimica:
it's nearly impossible to test against undefined JVM behavior of MT memory access

PR #1424, opened 2025-08-16
fix duration denormalized error

### LINKED ISSUE #1401: Recording a Duration can result in java.lang.AssertionError: ... ms is denormalized error
Using a simple mockkng of a method that takes a `Duration` such as 

```kotlin
@Test
    fun `some mockk bug`() {
        class Something {
            fun takeADuration(v: Duration): Boolean {
                return false
            }
        }
        val mockSomething = mockk<Something>(relaxed = true)
        every { mockSomething.takeADuration(any()) } returns true

    }
```
occasionally results in an exception with a stack of:

```
SomeTest > initializationError FAILED
    java.lang.AssertionError: -160686050132 ms is denormalized
        at kotlin.time.Duration.constructor-impl(Duration.kt:49)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at kotlin.reflect.jvm.internal.calls.CallerImpl$Method.callMethod(CallerImpl.kt:97)
        at kotlin.reflect.jvm.internal.calls.CallerImpl$Method$Static.call(CallerImpl.kt:106)
        at kotlin.reflect.jvm.internal.calls.ValueClassAwareCaller.call(ValueClassAwareCaller.kt:204)
        at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:108)
        at io.mockk.impl.recording.JvmSignatureValueGenerator.signatureValue(JvmSignatureValueGenerator.kt:24)
        at io.mockk.impl.recording.states.RecordingState.matcher(RecordingState.kt:46)
        at io.mockk.impl.recording.CommonCallRecorder.matcher(CommonCallRecorder.kt:52)
...
        at io.mockk.impl.eval.RecordedBlockEvaluator$record$block$1.invoke(RecordedBlockEvaluator.kt:24)
        at io.mockk.impl.eval.RecordedBlockEvaluator$enhanceWithRethrow$1.invoke(RecordedBlockEvaluator.kt:76)
        at io.mockk.impl.recording.JvmAutoHinter.autoHint(JvmAutoHinter.kt:23)
        at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:52)
        at io.mockk.impl.eval.EveryBlockEvaluator.every(EveryBlockEvaluator.kt:30)
        at io.mockk.MockKDsl.internalEvery(API.kt:94)
        at io.mockk.MockKKt.every(MockK.kt:148)
...
```

I believe this is because in https://github.com/mockk/mockk/blob/b03c36f88ea974c1422962287f40bcb3b7250f1e/modules/mockk/src/jvmMain/kotlin/io/mockk/impl/recording/JvmSignatureValueGenerator.kt#L20 `cls.boxedClass` returns `Long` which just generates a random Long value at https://github.com/mockk/mockk/blob/b03c36f88ea974c1422962287f40bcb3b7250f1e/modules/mockk/src/jvmMain/kotlin/io/mockk/impl/recording/JvmSignatureValueGenerator.kt#L40.

However, The Duration [constructor](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/time/Duration.kt#L42) has constraints based on whether the LSB is set or not. So occasionally a random value is generated that doesn't meet these constraints and an assertion is thrown.

tigermint:
## Fix Duration constructor AssertionError in JvmSignatureValueGenerator

**Problem**: 
- Random Long generation could trigger Duration constructor AssertionError when values fell in the "denormalized" range [-MAX_NANOS_IN_MILLIS, MAX_NANOS_IN_MILLIS].

**Solution**: 
- Generate Long values in Duration's safe millisecond range, avoiding the denormalized range while maximizing value diversity.

**Changes**:
- Added `generateSafeLong()` method with proper range calculation
- Updated Long::class case to use safe generation  
- Added test coverage for denormalized range avoidance
- Updated API definitions for companion object

**Testing**: Verified generated values avoid AssertionError and stay outside denormalized range.

Fixes #1401

### DIFF
--- a/modules/mockk/src/jvmMain/kotlin/io/mockk/impl/recording/JvmSignatureValueGenerator.kt
+++ b/modules/mockk/src/jvmMain/kotlin/io/mockk/impl/recording/JvmSignatureValueGenerator.kt
@@ -70,19 +70,20 @@ class JvmSignatureValueGenerator(val rnd: Random) : SignatureValueGenerator {
      */
     private fun generateSafeLong(): Long {
         return if (rnd.nextBoolean()) {
-            // Generate positive value in safe millisecond range (avoids denormalized)
+            // Generate positive value in safe millisecond range
             rnd.nextLong(MAX_NANOS_IN_MILLIS + 1, MAX_MILLIS)
         } else {
-            // Generate negative value in safe millisecond range (avoids denormalized)
+            // Generate negative value in safe millisecond range
             rnd.nextLong(-MAX_MILLIS, -MAX_NANOS_IN_MILLIS)
         }
     }
 
     companion object {
         // Duration internal constants (mirrored from kotlin.time.Duration)
+        // See: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/time/Duration.kt
         private const val NANOS_IN_MILLIS = 1_000_000L
         private const val MAX_NANOS = Long.MAX_VALUE / 2 / NANOS_IN_MILLIS * NANOS_IN_MILLIS - 1
         private const val MAX_MILLIS = Long.MAX_VALUE / 2
-        private const val MAX_NANOS_IN_MILLIS = MAX_NANOS / NANOS_IN_MILLIS
+        internal const val MAX_NANOS_IN_MILLIS = MAX_NANOS / NANOS_IN_MILLIS
     }
 }
--- a/modules/mockk/src/jvmTest/kotlin/io/mockk/impl/recording/JvmSignatureValueGeneratorTest.kt
+++ b/modules/mockk/src/jvmTest/kotlin/io/mockk/impl/recording/JvmSignatureValueGeneratorTest.kt
@@ -3,6 +3,7 @@ package io.mockk.impl.recording
 import io.mockk.impl.instantiation.AbstractInstantiator
 import io.mockk.impl.instantiation.AnyValueGenerator
 import io.mockk.impl.instantiation.CommonInstanceFactoryRegistry
+import io.mockk.impl.recording.JvmSignatureValueGenerator.Companion.MAX_NANOS_IN_MILLIS
 import kotlin.reflect.KClass
 import kotlin.test.Test
 import kotlin.test.assertTrue
@@ -31,12 +32,7 @@ class JvmSignatureValueGeneratorTest {
 
     @Test
     fun `generated Long values avoid denormalized range`() {
-        // Calculate denormalized range boundaries (same as implementation)
-        val nanosInMillis = 1_000_000L
-        val maxNanos = Long.MAX_VALUE / 2 / nanosInMillis * nanosInMillis - 1
-        val maxNanosInMillis = maxNanos / nanosInMillis
-
-        repeat(10) {
+        repeat(TEST_ITERATIONS) {
             val longValue = generator.signatureValue(
                 Long::class,
                 { mockAnyValueGenerator },
@@ -44,7 +40,7 @@ class JvmSignatureValueGeneratorTest {
             )
 
             assertTrue(
-                longValue !in -maxNanosInMillis..maxNanosInMillis,
+                longValue !in -MAX_NANOS_IN_MILLIS..MAX_NANOS_IN_MILLIS,
                 "Generated Long value $longValue is in denormalized range"
             )
         }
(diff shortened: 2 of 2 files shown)


modules/mockk/src/jvmTest/kotlin/io/mockk/impl/recording/JvmSignatureValueGeneratorTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test
+    fun `generated Long values avoid denormalized range`() {
+        // Calculate denormalized range boundaries (same as implementation)
+        val nanosInMillis = 1_000_000L
+        val maxNanos = Long.MAX_VALUE / 2 / nanosInMillis * nanosInMillis - 1
+        val maxNanosInMillis = maxNanos / nanosInMillis
+
+        repeat(10) {
```
Raibaz, line 39: I see what you're doing here and it makes sense to me.

I have a few possible improvements I can think of:
- I'd run each of these more than just 10 times, as the error in #1401 happens less frequently than 1 in 10 times.
- I'd extract the number of iterations in a constant and reuse it across all tests
- Can we run these iterations in parallel to speed up the test execution?

modules/mockk/src/jvmMain/kotlin/io/mockk/impl/recording/JvmSignatureValueGenerator.kt
```diff
@@ -62,4 +62,27 @@ class JvmSignatureValueGenerator(val rnd: Random) : SignatureValueGenerator {
...
+            rnd.nextLong(MAX_NANOS_IN_MILLIS + 1, MAX_MILLIS)
+        } else {
+            // Generate negative value in safe millisecond range (avoids denormalized)
+            rnd.nextLong(-MAX_MILLIS, -MAX_NANOS_IN_MILLIS)
+        }
+    }
+
+    companion object {
+        // Duration internal constants (mirrored from kotlin.time.Duration)
+        private const val NANOS_IN_MILLIS = 1_000_000L
```
Raibaz, line 83: Why do we mirror these? Can't we just import them from where they are defined?

modules/mockk/src/jvmMain/kotlin/io/mockk/impl/recording/JvmSignatureValueGenerator.kt
```diff
@@ -62,4 +62,27 @@ class JvmSignatureValueGenerator(val rnd: Random) : SignatureValueGenerator {
...
             }
     }
+
+    /**
+     * Generates a Long value in Duration's safe millisecond range to avoid AssertionError.
+     * Uses wider millisecond range while avoiding the denormalized range.
+     * See: https://github.com/mockk/mockk/issues/1401
+     */
+    private fun generateSafeLong(): Long {
+        return if (rnd.nextBoolean()) {
```
Raibaz, line 72: Why are we randomly generating a positive or a negative number?

Raibaz (changes requested):

Raibaz:
Thanks for looking into this! I left a few comments

modules/mockk/src/jvmMain/kotlin/io/mockk/impl/recording/JvmSignatureValueGenerator.kt
tigermint: They're not publicly accessible from kotlin.time.Duration. 😭

https://github.com/JetBrains/kotlin/blob/f2fa048b832148119b74e68d3cf89bf94760ad65/libraries/stdlib/src/kotlin/time/Duration.kt#L1037

modules/mockk/src/jvmMain/kotlin/io/mockk/impl/recording/JvmSignatureValueGenerator.kt
tigermint: To cover Duration's valid ranges. 

Duration has two safe ranges separated by the denormalized zone, 
so we randomly pick between the negative range `[-MAX_MILLIS, -MAX_NANOS_IN_MILLIS]` and positive range `[MAX_NANOS_IN_MILLIS + 1, MAX_MILLIS]`.

karsie:
Doesn't work on Android projects in instrumentation with mockk-android, since nextLong(Long, Long) doesn't exist

PR #1441, opened 2025-10-23
Add configurable logging to withArg & withNullableArg

OsaSoft:
## Problem decription:
After upgrading to a newer version I noticed a LOT of spamming in the console logs, enough to significantly slow down test execution and even crash the IDE. After some research I narrowed it down to this PR: https://github.com/mockk/mockk/pull/1395

The issue I encountered is caused by a combination of these factors:
- using property based testing (many generated test cases)
- verifying different calls to the same mock
- the change to `withArg` and `withNullableArg` dumping the error stack trace on any failed match, even if the test itself passes

## Proposed solution:
Add an argument flag to the `withArg` and `withNullableArg` functions to toggle the stacktrace dump behaviour

## Reproduction:
build.gradle.kts
```
import org.gradle.internal.impldep.org.junit.runner.Request.runner
import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType

plugins {
    kotlin("jvm") version "2.2.20"
}

group = "org.example"
version = "1.0-SNAPSHOT"

repositories {
    mavenCentral()
    mavenLocal()
}

dependencies {
    testImplementation(kotlin("test"))

    testImplementation("io.kotest:kotest-assertions-core:6.0.3")
    testImplementation("io.kotest:kotest-assertions-core-jvm:6.0.3")
    testImplementation("io.kotest:kotest-property:6.0.3")
    testImplementation("io.kotest:kotest-framework-engine-jvm:6.0.3")
    testImplementation("io.kotest:kotest-runner-junit5:6.0.3")

    testImplementation("io.mockk:mockk:1.14.7-SNAPSHOT-local")
}

tasks.test {
    useJUnitPlatform()
}
kotlin {
    jvmToolchain(21)
}
```
Foo.kt
```
package cloud.osasoft.mockkTest

interface Bar {
    fun doSomething(number: Int)
}

class Foo(private val bar: Bar) {

    fun fooBar(a: Int, b: Int) {
        bar.doSomething(a)
        bar.doSomething(b)
    }
}
```
FooSpec.kt
```
package cloud.osasoft.mockkTest

import io.kotest.assertions.shouldFail
import io.kotest.core.spec.style.FreeSpec
import io.kotest.matchers.ints.shouldBeLessThan
import io.kotest.matchers.shouldBe
import io.kotest.property.Arb
import io.kotest.property.arbitrary.int
import io.kotest.property.arbitrary.positiveInt
import io.kotest.property.checkAll
import io.mockk.mockk
import io.mockk.verify

class FooSpec : FreeSpec(
    {
        "This spams log a lot" {
            checkAll(
                Arb.int(1..10),
                Arb.int(1..10),
            ) { a, b ->
                val bar = mockk<Bar>(relaxed = true)
                val foo = Foo(bar)

                foo.fooBar(a, b)
                verify {
                    bar.doSomething(
                        withArg { it shouldBe a }
                    )
                    bar.doSomething(
                        withArg { it shouldBe b }
                    )
                }
            }
        }

        "This does not spam and successfully passes" {
            checkAll(
                Arb.int(),
                Arb.int(),
            ) { a, b ->
                val bar = mockk<Bar>(relaxed = true)
                val foo = Foo(bar)

                foo.fooBar(a, b)

                verify {
                    bar.doSomething(
                        withArg(logAllAssertionErrors = false) { it shouldBe a }
                    )
                    bar.doSomething(
                        withArg(logAllAssertionErrors = false) { it shouldBe b }
                    )
                }
            }
        }

        "This does not spam and fails" {
            checkAll(
                Arb.positiveInt(),
                Arb.positiveInt(),
            ) { a, b ->
                shouldFail {
                    val bar = mockk<Bar>(relaxed = true)
                    val foo = Foo(bar)

                    foo.fooBar(a, b)

                    verify {
                        bar.doSomething(
                            withArg(logAllAssertionErrors = false) { it shouldBe a }
                        )
                        bar.doSomething(
                            withArg(logAllAssertionErrors = false) { it shouldBe b }
                        )
                        bar.doSomething(
                            withArg(logAllAssertionErrors = false) {
                                // This will fail
                                it shouldBeLessThan 0
                            }
                        )
                    }
                }
            }
        }
    }
)
```

Here the first testcase spams the console with messages like this:
image

## Possible future improvements:
Would be nice if the behaviour could be configured project-wise via mockk config (would love some pointers on how to achieve this).
Would be nice if the stacktrace dump actually used a logger instead of directly dumping to console.

### DIFF
--- a/modules/mockk-dsl/api/mockk-dsl.api
+++ b/modules/mockk-dsl/api/mockk-dsl.api
@@ -292,13 +292,13 @@ public final class io/mockk/FunctionAnswer : io/mockk/Answer {
 }
 
 public final class io/mockk/FunctionMatcher : io/mockk/EquivalentMatcher, io/mockk/Matcher, io/mockk/TypedMatcher {
-	public fun <init> (Lkotlin/jvm/functions/Function1;Lkotlin/reflect/KClass;Z)V
-	public synthetic fun <init> (Lkotlin/jvm/functions/Function1;Lkotlin/reflect/KClass;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Lkotlin/jvm/functions/Function1;Lkotlin/reflect/KClass;Lkotlin/jvm/functions/Function1;)V
+	public synthetic fun <init> (Lkotlin/jvm/functions/Function1;Lkotlin/reflect/KClass;Lkotlin/jvm/functions/Function1;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun checkType (Ljava/lang/Object;)Z
 	public final fun component1 ()Lkotlin/jvm/functions/Function1;
 	public final fun component2 ()Lkotlin/reflect/KClass;
-	public final fun copy (Lkotlin/jvm/functions/Function1;Lkotlin/reflect/KClass;Z)Lio/mockk/FunctionMatcher;
-	public static synthetic fun copy$default (Lio/mockk/FunctionMatcher;Lkotlin/jvm/functions/Function1;Lkotlin/reflect/KClass;ZILjava/lang/Object;)Lio/mockk/FunctionMatcher;
+	public final fun copy (Lkotlin/jvm/functions/Function1;Lkotlin/reflect/KClass;Lkotlin/jvm/functions/Function1;)Lio/mockk/FunctionMatcher;
+	public static synthetic fun copy$default (Lio/mockk/FunctionMatcher;Lkotlin/jvm/functions/Function1;Lkotlin/reflect/KClass;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lio/mockk/FunctionMatcher;
 	public fun equals (Ljava/lang/Object;)Z
 	public fun equivalent ()Lio/mockk/Matcher;
 	public fun getArgumentType ()Lkotlin/reflect/KClass;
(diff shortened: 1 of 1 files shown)


Raibaz:
Thanks for the investigation!

What if instead of just passing a flag to turn on/off the `printStackTrace` calls we gave the option to pass a `java.util.logging.Logger` to use to log the stacktraces?

That would also solve at least one of the future improvements you identified.

OsaSoft:
Hi @Raibaz . Thats an interesting idea. How do you think we should handle the default there to maintain the current functionality? Nullable with default null, and if its null keep current behaviour, otherwise use the provided logger? Should we use a WARN or maybe even INFO level?

Or it could be something like this, that way when we dont want any logging, we can even completely skip the logger overhead, which for a lot of larger strings (like in this case) can actually have a noticable impact and we can leave it up to the user to use whatever log level they deem appropriate
`assertionErrorLogger: (AssertionError) -> Unit = { e -> e.printStackTrace() },`

then we could call the function e.g. like this:
```
withArg(assertionErrorLogger = logger::debug) {
  a shouldBe b
}
```

edit: actually that wouldnt work, since the receiver param would be the assertionError 🤔 would work with a `String` so then in the matcher we would have to call it with the exception converted to string, but that could lose the exception instance, if we wanted e.g. to log it like `logger.error("Failed matching arg", ex)`

Raibaz:
Uhm, why wouldn't it work?

We could implement the `FunctionMatcher` class like this:

```
data class FunctionMatcher<in T : Any>(
    val matchingFunc: (T) -> Boolean,
    override val argumentType: KClass<*>,
    private val assertionErrorLogger: (AssertionError) -> Unit = {e -> e.printStackTrace()}, // Existing behavior by default
) : Matcher<T>, TypedMatcher, EquivalentMatcher {
    override fun equivalent(): Matcher<Any> = ConstantMatcher(true)

    override fun match(arg: T?): Boolean {
        return if (arg == null) {
            false
        } else {
            try {
                matchingFunc(arg)
            } catch (a: AssertionError) {
                assertionErrorLogger();
                false
            }
        }
    }
```

And change the `withArg` function to accept a `(AssertionError) -> Unit` param to be relayed to the FunctionMatcher.

I quite like it, as it offer good customizability; I actually like it way more than just passing a logger.

OsaSoft:
i meant that a very short and clean reference pass wouldnt work, ie `withArg(assertionErrorLogger = logger::debug)`, but yeah, with lambda bodies it should work fine. Let me do a few quick tests, see what works best for a nice syntax and ill get back to you once i push the changes

OsaSoft:
ok, so i think this signature works quite well, some example usages
```
// quiet
withArg({}) { it shouldBe a }

// logger
withArg({ logger.debug(it) }) { it shouldBe a}

// custom reference
withArg(::logAssertion) { it shouldBe a }
fun logAssertion(error: AssertionError) = ...
```


one more thing Im considering: since right now the default function body is defined the same in quite a few places, could it be moved to possibly some static value? Where could this go, is there some file containing different util functions etc? Or maybe some config class that could be a good place to put it?

Raibaz:
Can you please run `gradle :mockk-dsl:apiDump` to update the api definition files and make the builds pass?

OsaSoft:
ok, should be done. Thanks for letting me know which exactly to run :)

Raibaz:
Thanks!

hho:
Maybe I don't completely understand what's going on here, but I believe such a breaking API change shouldn't be hidden in a patch release?

PR #1442, opened 2025-10-25
Fix #1342: Handle value classes for type parameters and don't unbox value classes returned as interface/supertype

### LINKED ISSUE #1342: ClassCastException when stubbing a method that returns T where T is a value class
> Unfortunately, there is still an issue with this feature. If you have an instance that extends from a class with generic values, if the output value of an execution of a function is a value class instance a `ClassCastException` is thrown:
> 
> In this case `ProductToProductInformationUiModelMapper` extends from:
> 
> ```kotlin
> 
> abstract class Mapper<in T, out R> {
>     suspend fun map(parameter: T) : R
> }
> 
> ```
> 
> In this case `R` will be:

> ```kotlin
> value class ProductItemUiModel(value: String) {
> // Some value class validation
> }
> 
> ```
>
> ```kotlin
> 
> private val productToProductInformationUiModelMapper: ProductToProductInformationUiModelMapper = mockk()
> 
> private val thingToTest: ThingToTest = ThingToTest(productToProductInformationUiModelMapper)
>
> @Test
> fun `test somthing`() {
>     val expectedValue = getSomeExpectedValue()
>     val productItemUiModelValue = expectedValue.productItemUiModelValue
>     val product = generateRandomProduct()
> 
>     coEvety { productToProductInformationUiModelMapper.map(product) } returns 
>     ProductItemUiModel(productItemUiModelValue)
> 
>     runTest {
>          val actualValue = thintToTest.doSomethingCool(product)
>     }
> }
> ```
> 
> Running this test throws:
> 
> ```
> class java.lang.String cannot be cast to class com.some.package.ProductInformationUiModel
> ```
> 

 _Originally posted by @j1c1m1b1 in [#152](https://github.com/mockk/mockk/issues/152#issuecomment-1737240277)_

ianbrandt:
Handle generic type parameters of value classes. Don't unbox value classes that are returned as their interface or supertype to avoid a `ClassCastException`. See added KDoc for details. Fixes mockk#1342.

### DIFF
--- a/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
+++ b/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
@@ -1,8 +1,6 @@
 package io.mockk.core
 
 import io.mockk.core.ValueClassSupport.boxedValue
-import java.lang.invoke.MethodHandle
-import java.lang.invoke.MethodHandles
 import java.lang.reflect.Method
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
@@ -48,12 +46,12 @@ actual object ValueClassSupport {
                 else -> return this
             }
 
-        // For generic return types, use a cached MethodHandle to probe for null.
+        // For generic return types, use a cached Method to probe for null.
         // This avoids calling mocked/intercepted property getters.
         if (expectedReturnType !is KClass<*>) {
-            val handle = resultType.resolveUnboxHandleOrNull()
-            if (handle != null) {
-                handle.invoke(this) ?: return null
+            val method = resultType.resolveUnboxMethodOrNull()
+            if (method != null) {
+                method.invoke(this) ?: return null
             }
             return this
         }
@@ -149,23 +147,18 @@ actual object ValueClassSupport {
     private val valueClassFieldCache = mutableMapOf<KClass<out Any>, KProperty1<out Any, *>>()
 
     /**
-     * Cached [MethodHandle]s to synthetic `unbox-impl` methods.
+     * Cached [Method]s to synthetic `unbox-impl` methods.
      */
-    private val valueClassUnboxHandleCache = mutableMapOf<KClass<out Any>, MethodHandle?>()
+    private val valueClassUnboxMethodCache = mutableMapOf<KClass<out Any>, Method?>()
 
     /**
-     * Lookup used to unreflect the discovered method into a MethodHandle (lazily created).
+     * Resolves a [Method] for the `unbox-impl` method of a value class.
      */
-    private val methodHandleLookup by lazy { MethodHandles.lookup() }
-
-    /**
-     * Resolves a [MethodHandle] for the `unbox-impl` method of a value class.
-     */
-    private fun <T : Any> KClass<T>.resolveUnboxHandleOrNull(): MethodHandle? {
+    private fun <T : Any> KClass<T>.resolveUnboxMethodOrNull(): Method? {
         // There's no such method for non-value classes
         if (!this.isValue_safe) return null
 
-        return valueClassUnboxHandleCache.getOrPut(this) {
+        return valueClassUnboxMethodCache.getOrPut(this) {
             try {
                 // Find zero-arg Kotlin synthetic unbox method: *unbox-impl
                 val unboxImplMethod =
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -6,6 +6,8 @@ build/
 
 .gradle/
 
+.kotlin/
```
ianbrandt: Created by Kotlin Gradle Plugin during tests.

test-modules/client-tests/src/jvmTest/kotlin/io/mockk/core/ValueClassSupportTest.kt
```diff
@@ -20,8 +22,345 @@ class ValueClassSupportTest {
         assertEquals(JavaEnum.A, result)
     }
 
+    @Test
```
ianbrandt: Added several test cases covering value class handling.

ianbrandt:
I see there was a build failure:

https://github.com/mockk/mockk/actions/runs/18830786451/job/53742959648?pr=1442#step:9:630

I'm not particularly familiar with Android development. Is this something I need to fix in my PR, or an issue with the build infrastructure?

Raibaz:
@ianbrandt , can you please rebase on top of the master branch? I should have fixed the build failure.

ianbrandt:
> can you please rebase on top of the master branch? I should have fixed the build failure.

@Raibaz, Done. Thank you for considering this PR.

Additional notes:

As I dug into this with test cases, I realized the issue I described in https://github.com/mockk/mockk/issues/1342#issuecomment-2655121254 turned out to be a separate, much simpler to fix bug than the generics issue originally reported by @micheljung. I’ve attempted to fix both in this PR.

Admittedly, not being deeply familiar with MockK or Kotlin value class internals, I leaned on [Junie](https://www.jetbrains.com/junie/) quite a bit to help develop the code and documentation for the cached "*unbox-impl" `MethodHandle` approach that ultimately got the generic value class tests passing. I did review and refine it, so it’s not just blindly used AI-generated code. That said, I’m not sure whether it's the best possible solution. Feedback welcomed. If there's a more recommended approach regarding MockK’s architecture or long-term Kotlin value class compatibility, I’m happy to take stab at that. I'm also on the fence about the added KDoc and comments. They do explain what's going on, but they're also a bit much. If you want me to pare them down, I can.

Raibaz:
Thanks for this! I like the test coverage particularly.

I agree the KDoc and comments in ValueClassSupport.kt are a bit too much, can you clean them up a bit?

ianbrandt:
> I agree the KDoc and comments in ValueClassSupport.kt are a bit too much, can you clean them up a bit?

You got it. Will do.

ianbrandt:
@Raibaz, I finally got a chance to clean up the KDoc and inline comments for this change. I also rebased the changes onto the latest master (hence the force push to this branch). Let me know if you see anything else that needs fixing.

Raibaz:
Can you also please run `./gradlew spotlessApply` to make the builds pass? 🙏 

Thanks!

ianbrandt:
Done. I wasn't familiar with Spotless. I'll be sure to run that after any other changes going forward. Thanks!

Raibaz:
Looks like there are some test failures in older Kotlin versions and in instrumented tests. Can you PTAL?

Thanks! 🙏

ianbrandt:
> Looks like there are some test failures in older Kotlin versions and in instrumented tests. Can you PTAL?

I believe the client test failures with earlier versions of Kotlin were due to them being run with the latest Kotlinx Coroutines version, which is not that backwards compatible. I believe any `coEvery` in `:test-modules:client-tests` would've failed, there just wasn't any such coverage previously. I changed the build to use the latest compatible Kotlinx Coroutines version for each Kotlin version. The corresponding `coroutinesVersion` logic in 'test-modules/client-tests/build.gradle.kts' may be a point of maintenance going forward.

I believe the instrumented tests were failing because they require compatibility with API level 21, and I had used `java.lang.invoke.MethodHandle`, which only became available from API level 26. I changed to using `java.lang.reflect.Method` instead. All the `check` tests still pass. I'm not familiar with how to setup to run `connectedCheck` locally yet. I'll be watching for whether they pass if you approve the next workflow run.

P.S. I also rebased against current master and checked `spotlessApply`.

Raibaz:
Thanks a lot for all the follow ups, this looks very clean now! 🙏

Issue #818, opened 2022-05-04
Feature Request - Suppress & WhiteBox like feature in mockK


srinivas175:
- [x] I am running the latest 

- [x] I checked the documentation and found no 

- [x] I checked to make sure that this issue has not already been filed

##### `Suppress unwanted behaviour`
I'm using the MockK library to write my kotlin android unit tests and it's really pain in my ass while writing unit tests for Android activity classes which have override methods like `onCreate()`, `onResume()`, `onNewIntent()` etc that call their parent classes. 

I'm looking for a feature in mockk with which i can suppress the parent class calls and complete my unit tests peacefully. Currently I've to write hybrid unit tests by using PowerMockito along with mockK library to use the [suppress functionality](https://github.com/powermock/powermock/wiki/Suppress-Unwanted-Behavior) provided by PowerMockito, due to lack of such cool feature or function in mockK.

#### `WhiteBox`
[WhiteBox](https://www.javadoc.io/doc/org.powermock/powermock-reflect/1.6.4/org/powermock/reflect/Whitebox.html) is another useful one from PowerMockito which is missing in mockK and would really be helpful if available on this library.

Current Specs:
* MockK version: V1.12.3 
* OS: Mac OS
* Kotlin version: 1.6.0

ayu-111:
@Raibaz If this is still up for grabs and beginner friendly issue, can you assign it to me please?

Raibaz:
Sure, go ahead!

ch200203:
I noticed that this issue has been open for quite some time and seems to be unassigned for a while.  
If it's still available, I’d be happy to look into implementing this feature (both the `suppress` and `whitebox`-like functionality).  

Before starting, could you please confirm if it’s okay for me to proceed?  
Also, any hints or guidance on how you’d like this feature to be designed or tested within the existing MockK architecture would be greatly appreciated.

Raibaz:
Sure, feel free to start working on this!

IIUC, what the `suppress` feature does is explicitly stubbing a function, like a superclass constructor, to ensure it's never called.

If that's the case, then `suppress` would just be an alias for `mockk` and we don't really need it :)
I'd start by confirming this hypothesis, and, if it's correct, then we could just add a mention in the docs describing how to obtain the same behavior as Powermockito's suppress in MockK.

ch200203:
@Raibaz Thanks for your guidance on this issue.
[...]

PR #1444, opened 2025-10-31
docs(readme): document suppressing superclass calls

ch200203:
### Summary

Added a short guide to the **Features** section of the README explaining  
how to suppress superclass method calls using existing MockK features (`spyk` + `every { ... } just runs`).

This provides a simple, built-in alternative to PowerMockito’s `suppress` functionality,  
showing how to isolate subclass logic during testing without executing unwanted parent behavior.

---

### Motivation

Users migrating from PowerMockito or testing complex class hierarchies (such as framework base classes)  
often look for a way to “suppress” `super` method calls.  
MockK already supports this pattern, but it was not explicitly documented.

Adding this example improves discoverability and helps users understand how to achieve  
the same behavior using `spyk` and explicit stubbing.

---

### Changes

- Updated `README.md`  
  - Added a new subsection **“Suppressing superclass calls”** under **Features**
  - Included an example using a simple `Parent` / `Child` class hierarchy  
    demonstrating suppression via `every { ... } just runs`
  - Added a short explanation for users coming from PowerMockito

---

### Notes

- Documentation-only change — no code modifications.
- Based on discussion in #818

### DIFF
--- a/README.md
+++ b/README.md
@@ -1241,6 +1241,7 @@ To suppress a method call, especially a `super` call inside an overridden method
 Using `every { ... } just runs` replaces the entire method body, preventing the original code from executing.
 
 This is particularly useful for users coming from frameworks like PowerMockito or for testing classes like Android Activities.
+For reference, see [PowerMockito suppress documentation](https://github.com/powermock/powermock/wiki/Suppress-Unwanted-Behavior).
 
 ```kotlin
 // A simple inheritance hierarchy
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1235,6 +1235,42 @@ mockkObject(ObjBeingMocked) {
 }
 ```
 
+### Suppressing superclass calls
+
+To suppress a method call, especially a `super` call inside an overridden method, you can stub its behavior on a `spyk`.
+Using `every { ... } just runs` replaces the entire method body, preventing the original code from executing.
+
+This is particularly useful for users coming from frameworks like PowerMockito or for testing classes like Android Activities.
```
Raibaz: How about adding a link to the documentation for PowerMockito's suppress here? (https://github.com/powermock/powermock/wiki/Suppress-Unwanted-Behavior)

Raibaz (changes requested):

ch200203:
Thanks @Raibaz! 🙏  
Added the full stop and PowerMockito suppress link as suggested 
Let me know if everything looks good now.

PR #1448, opened 2025-11-08
Centralizing dependencies, bumping dependency versions

jgrnrt:
changes:
- Gradle 9.2.1
- switch to libs.versions.toml
- kotlin 2.2.21
- coroutines 1.10.2
- logback 1.5.20
...

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -52,9 +52,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        api-level: [ 30 ] # Relevant SDK versions to test for MockK Android projects
+        api-level: [ 26, 30, 36 ] # Relevant SDK versions to test for MockK Android projects
       fail-fast: false # in case one API-level fails, we still want to see results from others
-    name: "[api-level=${{ matrix.api-level }}]"
+    name: "android: [api-level=${{ matrix.api-level }}]"
     timeout-minutes: 30
     steps:
       - name: checkout
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -15,29 +15,29 @@ jreleaser = "1.21.0"
 
 # Logging
 slf4j = "2.0.17"
-logback = "1.5.20"
+logback = "1.5.21"
 
-byte-buddy = "1.15.11"
-objenesis = "3.3"
-dexmaker = "2.28.3"
+byte-buddy = "1.18.2"
+objenesis = "3.4"
+dexmaker = "2.28.6"
 
 junit4 = "4.13.2"
 junit = "5.14.1"
 
 kotlinx-benchmark = "0.4.14"
 
-android-gradle = "8.13.0"
+android-gradle = "8.13.1"
 
 # Android SDKs
 min-sdk = "26"
-sdk = "34"
+sdk = "36"
 
 # Android Test
-androidx-espresso = "3.5.1"
-androidx-orchestrator = "1.4.2"
-androidx-rules = "1.5.0"
-androidx-runner = "1.5.2"
-androidx-junit = "1.1.5"
+androidx-espresso = "3.7.0"
+androidx-orchestrator = "1.6.1"
+androidx-rules = "1.7.0"
+androidx-runner = "1.7.0"
+androidx-junit = "1.3.0"
 
 [libraries]
 kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
--- a/modules/mockk/build.gradle.kts
+++ b/modules/mockk/build.gradle.kts
@@ -36,14 +36,16 @@ kotlin {
             dependencies {
                 compileOnly(libs.slf4j)
 
-                implementation(libs.junit4)
+                compileOnly(libs.junit4)
+
                 implementation(dependencies.platform(libs.junit.bom))
-                implementation("org.junit.jupiter:junit-jupiter")
-                runtimeOnly("org.junit.platform:junit-platform-launcher")
+                compileOnly("org.junit.jupiter:junit-jupiter")
             }
         }
         val jvmTest by getting {
             dependencies {
+                implementation(libs.junit4)
+
                 implementation(dependencies.platform(libs.junit.bom))
                 implementation("org.junit.jupiter:junit-jupiter")
                 runtimeOnly("org.junit.platform:junit-platform-launcher")


jgrnrt:
I've made some necessary changes to bump to gradle 9. Somehow the tests now need more heap space. I've increased the heap space for tests. 
One test still has the error "Caused by: java.lang.OutOfMemoryError: Java heap space" at ProxyWeakReferenceTest.kt:12.
I've already tried bumping the maxHeapSize to 12g (for testing), but that still didn't work.
Could someone have a look at that?

Raibaz:
The ProxyWeakReferenceTest is indeed stressing the memory to verify that there are no leaks; it surprises me that upgrading dependencies makes it throw OOME.

I suggest adding `-XX:+HeapDumpOnOutOfMemoryError` to the jvm args and trying to understand which dependency upgrade caused the failure.

jgrnrt:
I cloned master version of the project in another directory and ran some tests.
I experience weird behavior, which may have been fixed with gradle 9.

Output with Gradle 8:
```
./gradlew jvmTest --rerun-tasks
> Task :modules:mockk:jvmTest FAILED
[...]
* What went wrong:
Execution failed for task ':modules:mockk:jvmTest'.
[...]
BUILD FAILED in 1m 1s
47 actionable tasks: 47 executed
```
Same java.lang.OutOfMemoryError: Java heap space in ProxyWeakReferenceTest.kt:12 with debug logs.

Running
```
./gradlew check --rerun-tasks
[...]
ProxyWeakReferenceTest[jvm] > test()[jvm] SKIPPED
There were failing tests. See the report at: file:///home/jere/Projects/mockk-clean/mockk/modules/mockk/build/reports/tests/jvmTest/index.html

[...]
BUILD SUCCESSFUL in 1m 2s
```
And running with debug shows the heap error as well:
```
 Caused by:
[TestEventLogger]         java.lang.OutOfMemoryError: Java heap space
[TestEventLogger]             at io.mockk.it.ProxyWeakReferenceTest.test(ProxyWeakReferenceTest.kt:12)
[TestEventLogger] Gradle Test Run :modules:mockk:jvmTest[jvm] FAILED
[org.gradle.internal.buildevents.BuildResultLogger] BUILD SUCCESSFUL in 1m 7s
```

Gradle 9 may have fixed this behavior, so gradle check might now correctly show BUILD FAILED. 
It would be helpful if someone else could confirm this behavior with gradle 8.

Additionally, my computer crashed several times before I realized that rerunning those tests were filling up my memory. 
Running ./gradlew --stop between test runs prevents this from happening.

The ProxyWeakReferenceTest is leaking memory. Or at least the created objects are not being removed from GC, which can be seen in heap dump. I am currently looking into it, but maybe someone with more experience with the project could have a look as well.

Raibaz:
Can you please rebase on top of master and retry?

I _think_ I found the cause of the memory leak spotted by `ProxyWeakReferenceTest` and fixed it.

Raibaz:
...nevermind, that didn't work.

I'll keep looking into a fix.

jgrnrt:
## Test problems
ProxyWeakReferenceTest prevents subsequent tests from running across all branches. Disabling this problematic test allows the remaining tests to execute properly.
Additionally, we can now see that VerificationAcknowledgeTest has been failing for some time. The two failing tests in VerificationAcknowledgeTest only pass when run in isolation.

## Changes
Centralized all dependencies into libs.versions.toml for easier dependency management
Fixed several deprecation warnings.
Check commit messages for details.

## JUnit 6
How should we proceed with the JUnit 6 dependency? Options:

- Revert to JUnit 5.14.1 to maintain current Java and Kotlin baseline compatibility
- Continue with JUnit 6 and accept the updated baseline requirements

I'm not certain postponing the JUnit 6 migration is the best approach. What are your thoughts?

.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
```
Raibaz: Was it intentional to commit this?

modules/mockk/src/jvmTest/kotlin/io/mockk/it/ProxyWeakReferenceTest.kt
```diff
@@ -3,20 +3,23 @@ package io.mockk.it
 import io.mockk.clearMocks
 import io.mockk.every
 import io.mockk.spyk
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 
 class ProxyWeakReferenceTest {
+    @Disabled("Leaks memory. Objects are not removed from GC. https://github.com/mockk/mockk/pull/1448 TBD: leak issue")
```
Raibaz: I don't think we should disable this. It is meant to verify that MockK doesn't leak memory, if we disable we are explicitly accepting that MockK has a memory leak.

Raibaz:
> How should we proceed with the JUnit 6 dependency? Options:
> 
> * Revert to JUnit 5.14.1 to maintain current Java and Kotlin baseline compatibility
> * Continue with JUnit 6 and accept the updated baseline requirements
> 
> I'm not certain postponing the JUnit 6 migration is the best approach. What are your thoughts?

Given that JUnit 6 was released less than two months ago and the vast majority of the existing codebases still rely on JUnit 5, I think we should hold off on breaking compatibility with JDK <17 and Kotlin <2 for now.

jgrnrt:
Only one additional test was running endlessly. I've disabled it.

Project builds very quick now with activated configuration cache.

Dokka generation for the whole project works as well. Split package links can't be resolved with Dokka. Which hopefully will be fixed in the future. See https://github.com/Kotlin/dokka/issues/4364

jgrnrt:
I am not sure on how to fix the illegal class access error:

```
io.mockk.MethodDescriptionTest > test[Medium_Phone_API_36.1(AVD) - 16] FAILED 
        java.lang.IllegalAccessError: Illegal class access ('kotlin.jvm.internal.CallableReference' attempting to access 'kotlin.jvm.internal.CallableReference$NoReceiver') in attempt to invoke static method kotlin.jvm.internal.CallableReference$NoReceiver kotlin.jvm.internal.CallableReference$NoReceiver.access$000() (declaration of 'kotlin.jvm.internal.CallableReference' appears in /data/app/~~rdGCj-nzuoij6zmLmUwS4w==/io.mockk.android.test-Z2ccNMDpWYiNX1D14Y-beA==/base.apk)
        at kotlin.jvm.internal.CallableReference.<clinit>(CallableReference.java:48)
```

Issue #1103, opened 2023-06-22
Problem when mocking a function with optional value class


sdetilly:
It seems that Mockk has a problem mocking a function that returns an optional value class. Here is an example:

```kotlin
    @JvmInline
    value class ValueTestClass(val int: Int)

    interface ClassProvider {
        fun provideTestClass(): ValueTestClass?
    }

    @Test
    fun someTest() {
        val mock = mockk<ClassProvider> {
            every { provideTestClass() } returns ValueTestClass(4)
        }

        assertEquals(ValueTestClass(4), mock.provideTestClass())
    }
```
Because `provideTestClass()` returns a nullable value class, the test will hang until I receive a `java.lang.OutOfMemoryError: Java heap space`. This happens even if I return a non-null value with the `every ... returns` block

When using breakpoints, it seems that the code is stuck in a loop inside the `every {}` block, but I don't quite understand why.

If I modify the interface so that `provideTestClass()` returns a non-null value class, the test will pass without issue

* MockK version: 1.13.5
* OS: MacOS 13.1
* Kotlin version: 1.8.21
* JDK version: JDK 18 (also fails with JDK 17)
* JUnit version: 4
* Type of test: unit test

VasilisDrettas-tomtom:
This seems similar to the issue I fixed here https://github.com/mockk/mockk/issues/813.

ncomet:
It's still relevant @VasilisDrettas-tomtom

I still can reproduce with:

mockk 1.13.13
jdk 21  Corretto-21.0.5.11.1
Kotlin 1.9.20

sdetilly:
We can close this. PR 1449 fixes this issue

PR #1449, opened 2025-11-19
Fix for issue #1103.

sdetilly:
The issue here was that `innermostBoxedClass` is recursive and was causing an infinite loop for nullable value classes, which are not inlined in JVM bytecode (they remain boxed), so it kept trying to unwrap something that shouldn't be unwrapped.

Also added a check to skip the boxedClass call entirely for nullable return types, since nullable value classes are not inlined and should be treated as regular boxed types.

Added 2 tests for this usecase and all tests pass using `gradle check`

Link to issue: https://github.com/mockk/mockk/issues/1103

### DIFF
--- a/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
+++ b/modules/mockk-core/src/jvmMain/kotlin/io/mockk/core/ValueClassSupport.kt
@@ -34,8 +34,8 @@ actual object ValueClassSupport {
             //   method.kotlinFunction.returnType.classifier == Foo
             val expectedReturnType = kFunction.returnType.classifier
             val isReturnNullable = kFunction.returnType.isMarkedNullable
-            // Use boxedClass (one level) instead of innermostBoxedClass (recursive) to avoid infinite loops (issue #1103)
-            val isPrimitive = resultType.boxedClass.java.isPrimitive
+            // Use innermostBoxedClass with recursion limit to avoid infinite loops (issue #1103) while still handling nested value classes (issue #1308)
+            val isPrimitive = resultType.innermostBoxedClass().java.isPrimitive
             return if (
                 !(kFunction.isSuspend && isPrimitive) &&
                 resultType == expectedReturnType &&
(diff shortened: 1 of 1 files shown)


nishatoma:
I believe this is causing another issue now:

https://github.com/mockk/mockk/issues/1475


When I try to add a `||` in JvmMockFactoryHelper line 181 to accept nullable value classes then it passes the test cases mentioned in #1475 

```
/ For nullable value classes, we should NOT call boxedClass as they are not inlined in JVM bytecode
        // This prevents infinite loops in value class type resolution (issue #1103)
        val returnType: KClass<*> = if (returnTypeClass.isValue || !returnTypeNullable) returnTypeClass.boxedClass else returnTypeClass
```

Otherwise the current code in master will cause verification to fail this test case:

```
@Test
    fun `when mock with optional value then test passes`() {
        val mocked: Mocked = mockk()
        every { mocked.something(any()) } returns Id("bar")

        val tested = Tested(mocked)

        tested.test()

        verify { mocked.something(Id("foo")) }
    }
```

Where:

```
@JvmInline
        value class Id(val id: String)

        interface Mocked {
            fun something(data: Id): Id?
        }

        class Tested(
            private val mocked: Mocked
        ) {
            fun test() {
                val result = mocked.something(Id("foo"))
            }
        }
```

nishatoma:
Perhaps we just need to use the `innermostBoxedClass()` extension function that was introduced, within `JvmMockFactoryHelper ` then it wouldn't matter if the value class was nullable or not, @sdetilly what are your thoughts on this?

sdetilly:
I was able to reproduce the issue. The problem occurs during verification with exact value matching when the return type is nullable.

I've implemented a fix and will a open a new PR:
  - Made innermostBoxedClass public in ValueClassSupport.kt
  - Updated JvmMockFactoryHelper.kt to use innermostBoxedClass() for all cases (nullable and non-nullable)
  - Added your test case to ValueClassTest.kt and it now passes (So I don't think I broke anything)

The innermostBoxedClass approach is better because it has built-in recursion protection while still properly handling both nullable and nested value classes. This resolves issues #1103, #1308, and #1475 together.

All value class tests pass with this fix. Thank you for the issue report and suggested solution 🎉

PR #1454, opened 2025-11-29
Add optional restricted mock system property

nishatoma:
## Overview

Hello! 

Currently, we can enable restricted mocking via `mockk.properties` file. One thing that would be really helpful for large projects with multi-modules is the ability to have restricted mocking per class or even per gradle command via terminal. This would make it easier to tackle modules one-by-one, for example:

`./gradlew -Pmockk.throwExceptionOnBadMock=true feature:your-module:test`

I've added a couple of tests to demonstrate this works with current config as well.

@Raibaz, what are your thoughts on this?

### DIFF
--- a/modules/mockk/src/commonMain/kotlin/io/mockk/impl/restrict/RestrictMockkConfiguration.kt
+++ b/modules/mockk/src/commonMain/kotlin/io/mockk/impl/restrict/RestrictMockkConfiguration.kt
@@ -14,8 +14,10 @@ class RestrictMockkConfiguration(propertiesLoader: PropertiesLoader = DefaultPro
         userDefinedRestrictedTypes = loadRestrictedTypesFromConfig(properties)
 
         restrictedTypes = DEFAULT_RESTRICTED_CLAZZ + userDefinedRestrictedTypes
-        throwExceptionOnBadMock = loadThrowExceptionSetting(properties) ||
-            loadThrowExceptionSystemProperty()
+
+        // System Property takes precedence.
+        throwExceptionOnBadMock = loadThrowExceptionSystemProperty() ?:
+            loadThrowExceptionSetting(properties)
     }
 
     companion object {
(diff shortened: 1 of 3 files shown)


Raibaz:
Looks good to me.

Would you be to add a mention of this in the docs [here](https://mockk.io/#restricted-mocking-in-mockk)?

modules/mockk/src/commonMain/kotlin/io/mockk/impl/restrict/RestrictMockkConfiguration.kt
```diff
@@ -14,10 +14,12 @@ class RestrictMockkConfiguration(propertiesLoader: PropertiesLoader = DefaultPro
         userDefinedRestrictedTypes = loadRestrictedTypesFromConfig(properties)
 
         restrictedTypes = DEFAULT_RESTRICTED_CLAZZ + userDefinedRestrictedTypes
-        throwExceptionOnBadMock = loadThrowExceptionSetting(properties)
+        throwExceptionOnBadMock = loadThrowExceptionSetting(properties) ||
+            loadThrowExceptionSystemProperty()
```
TWiStErRob: I don't think `||` is the right combiner here, it should be `system ?: setting`.

Suppose the properties file contains:
`mockk.throwExceptionOnBadMock=true`
and you run
`gradlew -Pmockk.throwExceptionOnBadMock=false`
what would you expect?

I would run that command explicitly because "I just want to see the tests run without this check".

nishatoma:
> Looks good to me.
> 
> 
> 
> Would you be to add a mention of this in the docs [here](https://mockk.io/#restricted-mocking-in-mockk)?

Thank you for the feedback! I will do that once I get home:)

nishatoma:
@Raibaz @TWiStErRob Added MD and addressed comment, thank you!

PR #1474, opened 2025-12-10
Unify settings files by prioritizing mockk.properties in MockKSettings

### LINKED ISSUE #1468: Unify settings files
Currently there are two different settings files for Mockk:

1. `src/test/resources/io/mockk/settings.properties` to configure [settings](https://mockk.io/#settings-file) globally
2. `src/test/resources/mockk.properties` to configure [Restricted Mocking](https://mockk.io/#how-to-configure-restricted-mocking)

We should merge these files and provide a single configuration source, to reduce confusion and streamline configuration.

snowykte0426:
## Summary

  This PR addresses #1468 by unifying MockK's configuration files into a single location, which should help streamline the configuration experience for users.

## Problem

  Currently, MockK uses two separate configuration files:
  - io/mockk/settings.properties for MockKSettings (relaxed mode, stack traces, etc.)
  - mockk.properties for RestrictMockkConfiguration (restricted mocking settings)

  This dual-file approach has caused some confusion about where to configure different settings.

## Solution

  I've modified MockKSettings.kt to prioritize loading from mockk.properties at the classpath root, with a fallback to the legacy io/mockk/settings.properties
   location for backward compatibility.

## Changes

  - Updated MockKSettings initialization to:
    a. First attempt to load from mockk.properties (unified location)
    b. Fall back to io/mockk/settings.properties if not found (legacy support)
  - No changes needed for RestrictMockkConfiguration (already uses mockk.properties)

## Benefits

  - Simplified configuration: One file for all MockK settings
  - Reduced confusion: Clear, unified configuration location
  - Full backward compatibility: No breaking changes for existing users
  - Future-ready: Lays the groundwork for a potential deprecation path

## Testing

  The change maintains compatibility with existing test suites:
  - All settings properties remain accessible with the same keys
  - No property name conflicts (RestrictMockk settings use mockk. prefix)
  - Existing tests continue to pass without modification

## Future Considerations (Optional)

  If the maintainers think it would be valuable, I'd be happy to follow up with:
  - Documentation updates to guide users toward the unified mockk.properties approach
  - Deprecation warnings when loading from the legacy location
  - Any other improvements you'd suggest

  Please let me know if you'd like me to tackle any of these in this PR or in a follow-up!

  Fixes #1468

### DIFF
--- a/modules/mockk-core/src/jvmTest/kotlin/io/mockk/core/config/UnifiedPropertiesLoaderTest.kt
+++ b/modules/mockk-core/src/jvmTest/kotlin/io/mockk/core/config/UnifiedPropertiesLoaderTest.kt
@@ -39,7 +39,10 @@ class UnifiedPropertiesLoaderTest {
         deleteFromClasspath(UnifiedPropertiesLoader.LEGACY_PROPERTIES_FILE)
     }
 
-    private fun writeToClasspath(path: String, content: String) {
+    private fun writeToClasspath(
+        path: String,
+        content: String,
+    ) {
         resolveFromClasspath(path).run {
             parent?.let { Files.createDirectories(it) }
             Files.writeString(this, content)


polarene:
There is still one duplication remaining: `RestrictMockkConfiguration` uses `DefaultPropertiesLoader`, which is yet another mechanism for reading properties and duplicates the "mockk.properties" path. Maybe we should refactor MockKSettings or this one to unify the loading logic. I'm not familiar with multiplatform projects, and I can't understand if 'mockk-dsl' module depends on 'mockk', but we should try to remove duplication.

snowykte0426:
> There is still one duplication remaining: `RestrictMockkConfiguration` uses `DefaultPropertiesLoader`, which is yet another mechanism for reading properties and duplicates the "mockk.properties" path. Maybe we should refactor MockKSettings or this one to unify the loading logic. I'm not familiar with multiplatform projects, and I can't understand if 'mockk-dsl' module depends on 'mockk', but we should try to remove duplication.


 @polarene Thanks for pointing out the duplication in `DefaultPropertiesLoader`

  I started refactoring before seeing your comment, so I initially missed that path duplication. I've now updated the implementation to address the issues you
   raised.

  What's updated

  1. Introduced a unified loader
      • Added `UnifiedPropertiesLoader` in `mockk-core`
      • All property resolution goes through this implementation
      • `"mockk.properties"` is now defined in a single place

  2. Removed duplication across modules
      • `MockKSettings` now uses the unified loader
      • `DefaultPropertiesLoader` delegates to it (marked as `@Deprecated`)
      • `RestrictMockkConfiguration` picks it up automatically via the updated loader

  3. Module dependency rationale
      • `mockk-dsl` does not depend on `mockk`
      • Dependency chain: `mockk-core` ← `mockk-dsl` ← `mockk`
      • Placing the loader in `mockk-core` ensures both sides can use it without breaking boundaries

  4. Compatibility
      • Deprecated paths still work as fallbacks
      • No breaking changes for existing configurations

  All tests pass and the behavior matches previous versions.
  Let me know if you'd like this organized differently.

README.md
```diff
@@ -1619,37 +1647,26 @@ For example:
 
 ### How to Configure Restricted Mocking
 
-You can configure Restricted Mocking behavior using the `mockk.properties` file.
-
-#### 1. Creating the `mockk.properties` File
+You can configure Restricted Mocking behavior using the unified `mockk.properties` configuration file described in the [Configuration File](#configuration-file) section.
```
polarene: Same reason as above
```suggestion
You can configure Restricted Mocking behavior using the `mockk.properties` configuration file described in the [Configuration File](#configuration-file) section.
```

polarene:
Thanks @snowykte0426, I'm reviewing the latest changes.

snowykte0426:
Thank you for the detailed review! I've addressed all your feedback.

  ### Changes Made

  I refactored the implementation based on your suggestions:

  **UnifiedPropertiesLoader.kt:**
  - Fixed the path resolution issue by using absolute paths with leading `/`
  - Converted to `object` since the class is stateless
  - Simplified `loadProperties()` using idiomatic Kotlin (scope functions + Elvis operator)
  - Made constants `internal` for test access
  - Unified ClassLoader usage to prevent the broken fallback mechanism

  **UnifiedPropertiesLoaderTest.kt:**
  - Removed the AI-generated tests (you were right—those were weak assertions)
  - Added comprehensive tests for actual file loading scenarios
  - Switched to `kotlin.test.Test`
  - Implemented helper methods for dynamic resource file creation/deletion

  **README.md:**
  - Removed "unified" references (implementation detail users don't need)
  - Made paths more copy-paste friendly
  - Used link references to eliminate duplication
  - Strengthened deprecation language

  ### Additional Refactoring

  To fully eliminate the duplication you identified in `RestrictMockkConfiguration`, I also:

  **Removed deprecated files:**
  - Deleted `DefaultPropertiesLoader.kt` (was only delegating to `UnifiedPropertiesLoader`)
  - Removed the `PropertiesLoader` type alias in `io.mockk.impl.restrict.propertiesloader`
  - Updated `RestrictMockkConfiguration` to use `UnifiedPropertiesLoader` directly
  - Updated API declarations accordingly

  **Question:** I assumed these were internal-only since they were deprecated, but I realize `DefaultPropertiesLoader` was in the public API. Should I revert
  this removal and keep the deprecated classes for a transition period, or is it acceptable to remove them now since `UnifiedPropertiesLoader` provides the
  same functionality?

  ### Note on the Tests

  I apologize for not catching the AI-generated code in my initial implementation. I used an LLM to help with some parts and focused on the core logic, but I
  should have been more thorough in reviewing the test quality before pushing.

  The new tests are completely rewritten with proper file I/O testing.

  Let me know if any adjustments are needed!

snowykte0426:
I've updated the test implementation to use `kotlin.io.path` extensions as you suggested in your code review (commit: 778f2d5b). I replaced the
  `java.io.File` API with `kotlin.io.path` extensions (`writeText`, `deleteExisting`) and used `Files.createDirectories()` for parent directory creation.
  While attempting to use `createParentDirectories()`, I encountered an "Unresolved reference" error from the IDE and the build failed with:

  ```
  > Task :modules:mockk-core:compileTestKotlinJvm FAILED
  e: file:///Users/snowykte0426/Programming/mockk/modules/mockk-core/src/jvmTest/kotlin/io/mockk/core/config/UnifiedPropertiesLoaderTest.kt:49:13 Unresolved
  reference: createParentDirectories

  FAILURE: Build failed with an exception.

  * What went wrong:
  Execution failed for task ':modules:mockk-core:compileTestKotlinJvm'.
  > A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
     > Compilation error. See log for more details
  ```

  So I used the standard Java NIO `Files.createDirectories()` instead for that specific operation. All tests pass with this implementation and it now matches
  the approach you recommended. However, if there's a better solution or something I might have missed, please let me know.

Raibaz:
Can you please rebase on top of the `master` branch? It should be just a matter of formatting after I merged #1483

snowykte0426:
> Can you please rebase on top of the `master` branch? It should be just a matter of formatting after I merged #1483

 @Raibaz I've rebased on top of the latest master (58750d37) after #1483 merge. Resolved the Spotless formatting conflicts and applied all formatting
  rules. All commits are now cleanly rebased.

snowykte0426:
Fixed CI failures in fd9fbd5 - replaced `kotlin.io.path` extensions with Java NIO APIs.

  The tests were using `kotlin.io.path.writeText` and `deleteExisting`, which are only available in Kotlin 1.8.0+. However, CI runs tests against Kotlin
  1.7.22 (and 1.5.32), causing compilation errors.

  I switched to Java NIO APIs (`Files.writeString`, `Files.delete`) for backward compatibility, but I'm wondering - should I just keep using `kotlin.io.path`
  and let the older Kotlin version tests fail? Or is backward compatibility with Kotlin 1.7.22 important for this project?

  Let me know your preference and I can adjust accordingly.

Raibaz:
We have to maintain backwards compatibility with 1.7.*, yes.

polarene:
Hi! Catching up now, thanks for the updates. Let me review the changes again.

modules/mockk-core/src/jvmTest/kotlin/io/mockk/core/config/UnifiedPropertiesLoaderTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.StandardOpenOption
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class UnifiedPropertiesLoaderTest {
+    private val loader = UnifiedPropertiesLoader
```
polarene: This property is useless now that `UnifiedPropertiesLoader` is an object, consider inlining it.

polarene:
> **Question:** I assumed these were internal-only since they were deprecated, but I realize `DefaultPropertiesLoader` was in the public API. Should I revert this removal and keep the deprecated classes for a transition period, or is it acceptable to remove them now since `UnifiedPropertiesLoader` provides the same functionality?
> ### Note on the Tests
> 
@snowykte0426 thanks for your updates, everything seems good now. About `DefaultPropertiesLoader` I don't think it would be a problem removing it, despite being part of the public API it's not directly referenced by library users. @Raibaz  do you see any problems?

Sorry for the incompatible suggestions regarding kotlin.io.path. I didn't know about the 1.7 compatibility, I always use the latest Kotlin.

snowykte0426:
@polarene 
I’ve applied all five inline code suggestions in commit d784011:
- Inlined the loader property to remove an unnecessary intermediate variable
- Removed redundant StandardOpenOption parameters that are already defaults
- Extracted resolveFromClasspath() to eliminate duplication and improve maintainability
- Replaced the reference with the javaClass extension property

All tests are passing (4/4). Please let me know if you’d like any further adjustments.

snowykte0426:
CI failed: https://github.com/mockk/mockk/actions/runs/20718814279/job/59562044513?pr=1474

  The apiCheck failed because UnifiedPropertiesLoader was changed to an object, which modified the public API. Fixed by updating the API dump.([6d14427](https://github.com/mockk/mockk/pull/1474/commits/6d1442787e00bb806ad3d16b004060c928afbe96))

modules/mockk-core/src/jvmTest/kotlin/io/mockk/core/config/UnifiedPropertiesLoaderTest.kt
```diff
@@ -0,0 +1,60 @@
...
+        val config = loader.loadProperties()
+        assertEquals(config["name"], "default")
+        deleteFromClasspath(UnifiedPropertiesLoader.UNIFIED_PROPERTIES_FILE)
+        deleteFromClasspath(UnifiedPropertiesLoader.LEGACY_PROPERTIES_FILE)
+    }
+
+    private fun writeToClasspath(
+        path: String,
+        content: String,
+    ) {
```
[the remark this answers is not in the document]
polarene: Hello! Oh right, I didn't remember Raibaz introduced Spotless just recently. Sure, if the formatting is enforced with that style so be it, ignore my style suggestion.

Raibaz:
Thank  you both, @snowykte0426 and @polarene !

Issue #1473, opened 2025-12-07
1.14.7 raises minSdkVersion for Android


bddckr:
This is a seemingly undocumented change in that release.

It's breaking my projects that use minSdkVersion 25, giving me compile errors:

> Manifest merger failed : uses-sdk:minSdkVersion 25 cannot be smaller than version 26 declared in library [io.mockk:mockk-agent-android:1.14.7] /Users/chris/.gradle/caches/9.2.1/transforms/f6ee3cfdcd590d90fe216db6bafa9d63/transformed/mockk-agent-android-1.14.7/AndroidManifest.xml as the library might be using APIs not available in 25
>   	Suggestion: use a compatible library with a minSdk of at most 25,
>   		or increase this project's minSdk version to at least 26,
>   		or use tools:overrideLibrary="io.mockk.proxy.android" to force usage (may lead to runtime failures)

Is there a reason the minSdkVersion was raised? Can it be lowered back to where it was?

waffiqaziz:
Same issue here https://github.com/mockk/mockk/discussions/1471#discussion-9216588

bddckr:
@Raibaz Looks like it was your change in https://github.com/mockk/mockk/commit/daabf986cc1c11e13276c0c516d1f1874e681a4e. Can you share some details as to why that was done and whether it's really required? Thanks in advance!

wverdese:
Same here

Raibaz:
Fixed by #1477

Issue #1434, opened 2025-10-08
Support for JDK 25


SimonMarquis:
Mockk is currently not compatible with JDK 25.
It is very likely to be related to the current version of Byte Buddy who does not support ASM 9.8 yet:
https://github.com/mockk/mockk/blob/b08945962453f2a6236fe403b7a0001fb13aba6b/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt#L22

ASM 9.8 support was added in version 1.17.5: https://github.com/raphw/byte-buddy/releases/tag/byte-buddy-1.17.5

Unfortunately, it seems like Byte Buddy was previously downgraded in https://github.com/mockk/mockk/pull/1413 for compatibility issues. Is is still the case? And if so, could we schedule the deprecation/removal of old Android versions to be able to run this tool on modern codebases?

Raibaz:
Yep, the ByteBuddy version we depend on is still 1.15.11, but I'm definitely in favor of upgrading Android instrumentation to be able to use the latest version of ByteBuddy instead of downgrading it to comply with older Android versions.

peterdk:
When using
```
testImplementation("io.mockk:mockk:1.14.6")
testImplementation("net.bytebuddy:byte-buddy:1.17.8") // TEMP: Needed for mockk 1.14.6 with java25
```
my unittests succeed again on Java 25. (So for now useful as workaround)

jgrnrt:
Should be fixed in the latest version.

PR #1477, opened 2025-12-12
bump dependencies, bump dexter

jgrnrt:
Since I haven't found a solution for PR https://github.com/mockk/mockk/pull/1448 . I wanted to push some small changes and get all tests running again. (besides the disabled ones)

- added pipeline tests for several android sdks
- remove duplicate dependency
- disable not working tests
- bump kotlin 2.2.21
- gradle 9.2.1
- bump android gradle 8.13.2 and several androidx dependencies
- add latest dexter android-16.0.0_r4 and added some comments in readme
- remove idea plugin
- added gradle configuration-cache

### DIFF
--- a/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
+++ b/buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
@@ -29,7 +29,9 @@ object Deps {
 
         const val compileSdk = 34
         const val targetSdk = 34
-        const val minSdk = 21 // for static mocking is minSdk 28
+        const val minSdk = 21
+        // SDK 26+: Open classes, abstract classes, subclass proxies
+        // SDK 28+: Final classes/methods, static mocking
     }
 
     object Libs {


jgrnrt:
I've incorporated the JUnit Platform fix from my other PR and made several updates to the Gradle workflow:

Renamed some jobs for better clarity
Bumped several actions to their latest versions
Expanded from one to three Android Gradle jobs
Deactivated AVD caching due to a known issue

The AVD cache had to be disabled because running the emulator with caching enabled across multiple jobs causes them to hang (see https://github.com/ReactiveCircus/android-emulator-runner/issues/362).

Performance impact:

With cache: ~6 minutes per job
Without cache: 8-10 minutes per job

Given the stability issues with parallel jobs, I believe the 2-4 minute increase in build time per job is an acceptable trade-off for reliability.

buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
```diff
@@ -16,30 +16,29 @@ object Deps {
...
-        const val byteBuddy = "1.15.11"
-        const val objenesis = "3.3"
-        const val dexmaker = "2.28.3"
-        const val androidxEspresso = "3.5.1"
-        const val androidxOrchestrator = "1.4.2"
-        const val androidxTestRules = "1.5.0"
-        const val androidxTestRunner = "1.5.2"
-        const val androidxTestExtJunit = "1.1.5"
-        const val androidxTestOrchestrator = "1.4.2"
+        const val byteBuddy = "1.18.2"
```
Raibaz: This was intentionally downgraded in #1413, but it should be ok to re-upgrade now as discussed in #1434

buildSrc/src/main/kotlin/buildsrc/config/Deps.kt
```diff
@@ -16,30 +16,29 @@ object Deps {
...
+        const val androidxEspresso = "3.7.0"
+        const val androidxOrchestrator = "1.6.1"
+        const val androidxTestRules = "1.7.0"
+        const val androidxTestRunner = "1.7.0"
+        const val androidxTestExtJunit = "1.3.0"
 
         const val compileSdk = 34
         const val targetSdk = 34
-        const val minSdk = 26
+        const val minSdk = 28
```
Raibaz: Can we set this to 21 to fix #1473?

jgrnrt:
Added MinSdk JUnit Rule to exclude unsupported feature tests.

PR #1480, opened 2025-12-18
Fix issue 1475: nullable value class verification

sdetilly:
After merging https://github.com/mockk/mockk/pull/1449, we noticed a new issue involving JvmMockFactoryHelper, where verification does not work with nullable value classes.

We now use innermostBoxedClass() for all value classes (nullable and non-nullable). This function:
- Has built-in recursion protection (max depth of 10)
- Properly unwraps nested value classes
- Works correctly for both nullable and non-nullable types

All tests pass, so we hopefully don't introduce regressions this time.

Thanks to @nishatoma for the issue description and solution suggestion!


Raibaz:
Thanks for looking into this and the thorough explanation!

PR #1483, opened 2025-12-21
Add Spotless formatting plugin

jgrnrt:
Adds Spotless code formatting to the project via a new Gradle convention plugin and reformats existing code to match the style guidelines.

# Changes
- Added `spotless.gradle.kts` convention plugin
- Only Kotlin (`.kt`) and Kotlin Gradle (`.gradle.kts`) files are formatted using ktlint
- Reformatted 300+ files to comply with new formatting standards

I recommend installing the pre-push Git hook with `./gradlew spotlessInstallGitPrePushHook` to prevent pushing unformatted code.


Raibaz:
This is really huge.

Would it be possible to split it in a couple different PRs to ease review?

jgrnrt:
I know that it is huge, but splitting is not really feasible since the formatting changes touch so many files across the codebase.

The main issue is that Spotless will fail the Gradle check if files aren't formatted with spotlessApply. 

We have a few options:

1. Merge this as-is (my recommendation) - Get all formatting standardized in one go. Yes, it's a large diff, but it's mechanical changes that can be reviewed quickly by spot-checking a few files and trusting the Spotless configuration.
2. Use "ratchet" mode - Only format files as they're modified in future PRs. This avoids the big-bang change but means the codebase stays inconsistent for a long time, and you'll have mixed formatting in diffs for months/years.
3. Merge without review of every file - Since these are automated formatting changes, you could review the Spotless configuration itself and spot-check several files, then approve based on that.

I'd strongly recommend option 1. The pain is upfront but temporary, and then the pre-push hook prevents future formatting issues. Dragging it out makes the codebase messier longer.

Worth noting: The only functional change was updating one test that checked a line number - it shifted by +1 due to the formatting. Everything else is purely cosmetic.

What do you think?

Issue #1296, opened 2024-09-13
any<SpecificType>() matcher ignores the concrete type and matches all types


Zordid:
### Expected Behavior

When verifying calls, using `any<SpecificType>()` as an argument, it should only fit if the call happened with this `SpecificType` - for other types it should fail!
`inline fun <reified T : Any> any(): T` even has a more obvious sibling `fun <T : Any> any(classifier: KClass<T>): T`... problem/bug is the same here, the given `classifier` is completely ignored.

Interestingly, replacing `any<SpecificType>()` with `capture(slot<SpecificType>())` does this correctly - even though the capture slot is not really wanted or needed.

### Current Behavior

When you verify a call on a function `fun call(arg: Any)` using `verify { call(any<String>()) }`, it verifies OK for *any* argument type called, not only String type! 

#### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* MockK version: 1.13.12
* OS: MacOS Sonoma
* Kotlin version: 1.9.25
* JDK version: 11
* JUnit version: 5.9.1
* Type of test: unit test

#### Minimal reproducible code (the gist of this issue)

```kotlin
interface System {
    fun send(message: Any)
}

fun myCodeUnderTest(system: System) {
    system.send("foo")
}

class MockkDemo {

    @Test
    fun `this test should fail, but does not`() {
        val mockProducer = mockk<System>(relaxed = true)

        myCodeUnderTest(mockProducer)

        // I expect the following line to fail, because the message type for arg was a String, not an Int!
        verifySequence { mockProducer.send(any<Int>()) }
        // same here, even more obvious to me that the second argument should only match Int type
        verifySequence { mockProducer.send(any(Int::class)) }
    }

    @Test
    fun `using capture instead of any works correctly`() {
        val mockProducer = mockk<System>(relaxed = true)

        myCodeUnderTest(mockProducer)

        // this works just fine - it fails as it should!
        verifySequence { mockProducer.send(capture(slot<Int>())) }
    }

}
```

chapakook:
I'd like to work on this issue.
I'm adding a regression test and a fix. Could you assign this issue to me?

Raibaz:
Sure, go ahead!

chapakook:
Hello! 🙂 @Raibaz  I made #1494 as a draft PR — PTAL when you have time.
I added some context and questions in the PR description.

PR #1494, opened 2026-01-09
Fix type matching for any() when parameter is Any  (#1296)

chapakook:
Fixes: #1296

### Problem

The `any<T>()` and `any(KClass)` matchers do not respect the requested type when
a mocked method parameter is declared as `Any`.

As a result, verifications such as `any<Int>()` may incorrectly match arguments
of a different runtime type (e.g. `String`), leading to false-positive matches
during verification.

### Solution

Introduce a type-aware matcher (`AnyTypedMatcher`) for the `any` family of
matchers and apply type validation during invocation matching.

The solution:
- Preserves the type information provided to `any<T>()` and `any(KClass)`
- Ensures type checks are applied only when the runtime argument is non-null
- Supports Kotlin value classes whose runtime representation may differ from
  their declared type
- Keeps existing matcher behavior unchanged to avoid breaking compatibility

This fixes the incorrect matching behavior reported in #1296 while maintaining
existing semantics for other matchers.

### Changes

- Introduced `AnyTypedMatcher` to carry and validate type information for
  `any<T>()` and `any(KClass)`
- Refined `InvocationMatcher` to invoke `TypedMatcher.checkType` only for
  non-null arguments
- Added value-class–aware type handling scoped specifically to `AnyTypedMatcher`
- Added regression tests covering:
  - Incorrect matches for `any<T>()` and `any(KClass)` with `Any`-typed parameters
  - Correct matching for valid types
  - Comparison with `capture` behavior to ensure type safety

### Tests

Added `AnyMatcherTypeTest` to verify:
- `any<Int>()` and `any(Int::class)` do not match `String` arguments
- Correct matches for valid argument types
- Proper behavior when using `any(Any::class)`
- Consistency with `capture`-based type checks

### Note

Type-checking behavior for matchers other than `any()` remains unchanged.
Additional handling for Kotlin value classes is intentionally scoped to
`AnyTypedMatcher` to minimize side effects and preserve backward compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1466,7 +1466,7 @@ By default, simple arguments are matched using `eq()`
 
 | Matcher                                                 | Description                                                                                            |
 |---------------------------------------------------------|--------------------------------------------------------------------------------------------------------|
-| `any()`                                                 | matches any argument of type `T` (type-checked)                                                        |
+| `any<T>()`                                              | matches any argument of type `T` (type-checked)                                                        |
 | `any(Class)`                                            | matches any argument of the given Class (type-checked; useful for reflective mocking)                  |
 | `allAny()`                                              | special matcher that uses `any()` instead of `eq()` for matchers that are provided as simple arguments |
 | `isNull()`                                              | checks if the value is null                                                                            |


chapakook:
### What I changed

I updated the behavior of `any()` / `any(KClass)` matchers to improve type checking
when value classes and their boxed / underlying representations are involved.

Specifically:
- Added additional normalization logic in `AnyTypedMatcher.checkType`
- Extended matching to cover:
  - value class instances
  - boxed types
  - underlying boxed types inferred from value class constructors

The goal is to make `any<T>()` behave consistently when value classes are used
as arguments.

### Why this change is needed

With value classes, `any<T>()` may currently fail type checks depending on
how the argument is boxed at runtime.
This causes unexpected mismatches even though the argument is logically compatible.

This change attempts to make the matcher more permissive and predictable
in those scenarios.

### Current issue / question

During this change, `AnyTypedMatcher` became part of the public API of `mockk-dsl`,
which causes `jvmApiCheck` to fail.

Before proceeding further, I would like to confirm:

- Is it acceptable for `AnyTypedMatcher` to be public in `mockk-dsl`?
- Or should this matcher be kept internal / relocated to avoid an API surface change?

I'm happy to adjust the implementation based on your preferred direction.

Raibaz:
If it's a new matcher type, it's correct for it to be included in the public API.

You should be able to make the build pass by running `./gradlew apiDump`.

chapakook:
Thanks! 😄 Confirmed it's a new matcher type.
I ran `./gradlew apiDump` and committed the updated API snapshot (mockk-dsl.api).

Raibaz:
Thanks a lot for looking into this!

Would you also please add documentation about the new matcher type [here](https://mockk.io/#matchers)?

chapakook:
😄 Thanks for the suggestion. 
I’ve added documentation for the new matcher type.
Please let me know if anything else should be clarified.

Raibaz:
Thanks for following up on this! 🙏

Issue #1420, opened 2025-08-01
Gradual test slowdown when using clearAllMocks with method mock returning non primitive value


javolek:
We are observing massive gradual test slowdowns. The slowndown is more apparent if a lot of consecutive tests are run at once, with each further test getting slower. Running single test alone does not show any slowdown at all.

I have been able to strip down the Problem to tests meeting following conditions:
- test is mocking a method returning non-primitive value, i.e. `data class`
- test is performing `clearAllMocks` or `confirmVerified`

It looks like a kind of mocks leaks into further tests as the cause of the slowdown seems to be the execution of `clearAllMocks` or `confirmVerified` taking too much time.

Example:
```kotlin
@TestMethodOrder(MethodOrderer.OrderAnnotation::class)
class MainTest {

    val testRepository = mockk<TestRepository>()

    @RepeatedTest(5)
    @Order(1)
    fun fastTest() {
        measureTimeMillis {
            repeat(1000) { i ->
                every { testRepository.getInt(any()) } returns i
                assertEquals(i, testRepository.getInt(i))
                verify (exactly = 1) { testRepository.getInt(i) }
                clearAllMocks()
            }
        }.let { println(it) }
    }

    @RepeatedTest(5)
    @Order(2)
    fun slowTest() {
        measureTimeMillis {
            repeat(1000) { i ->
                every { testRepository.getTestValue(any()) } returns TestValue(i)
                assertEquals(TestValue(i),testRepository.getTestValue(i))
                verify(exactly = 1) { testRepository.getTestValue(i) }
                clearAllMocks()
            }
        }.let { println(it) }
    }
}

interface TestRepository {
    fun getTestValue(a: Int): TestValue
    fun getInt(a: Int): Int
}

data class TestValue(val id: Int)
```

The `fastTest` shows only a slight variance of the excution time, as one would expect, while the `slowTest` shows gradualy (almost exponentialy) increasing execution time. Also the order of the tests is important i if the `fastTest` is executed after the `slowTest` then it's excution time will be even longer then the longest exection of the `slowTest` but the execution time variance is moderate.

Thou there is a workarround to use `clearMocks` and specify the particular mocks instead of `clearAllMocks`, but as soon as the `confirmVerified` is executed with or without particular mocks the slowdown is back again.

- MockK Version: 1.14.5
- JDK Version: 21
- Kontlin Version: 2.1.21

nsk90:
[...]

PR #1503, opened 2026-01-18
Introduce clearAllStubsFromMemory

### LINKED ISSUE #1486: clearAllMocks() slows down test execution
We have a project with a very large number of modules and tests (each of them creates mocks).
We've encountered a problem with test slowdowns due to unexpected behaviour of the clearAllMock() function.

The library holds internal mock map, which should be cleared by GC. But in practice we don't ever see that it is actually cleared.

When clearAllMock() is called, the internal mock map is too large, causing each subsequent test to run progressively slower.

An example of this behavior:
> duration: 244 ms	 (map.count = 1001)
duration: 291 ms	 (map.count = 2001)
duration: 385 ms	 (map.count = 3001)
duration: 525 ms	 (map.count = 4001)
duration: 741 ms	 (map.count = 5001)
duration: 833 ms	 (map.count = 6001)
duration: 988 ms	 (map.count = 7001)
duration: 1063 ms	 (map.count = 8001)
duration: 1219 ms	 (map.count = 9001)
duration: 1485 ms	 (map.count = 10001)
duration: 1546 ms	 (map.count = 11001)
duration: 1815 ms	 (map.count = 12001)
duration: 2047 ms	 (map.count = 13001)
duration: 2445 ms	 (map.count = 14001)
duration: 2695 ms	 (map.count = 15001)
duration: 2964 ms	 (map.count = 16001)
duration: 3327 ms	 (map.count = 17001)
duration: 3824 ms	 (map.count = 18001)
.....

As quick a workaround we clear the internal library map via reflections manually.

> duration: 168 ms	 (map.count = 0)
duration: 68 ms	 (map.count = 0)
duration: 55 ms	 (map.count = 0)
duration: 52 ms	 (map.count = 0)
duration: 58 ms	 (map.count = 0)
duration: 43 ms	 (map.count = 0)
duration: 42 ms	 (map.count = 0)
duration: 45 ms	 (map.count = 0)
...

1) It would be nice to have explicit public function to clear the map, which could be called after each test manually.
2) We assume that GC map cleanup should work, but as we see in runtime it does not.

Would you accept a PR with such a public method to clear mock map if we open it?

I am attaching a file with a minimal example and a workaround.
[CarTest.kt.txt](https://github.com/user-attachments/files/24338369/CarTest.kt.txt)

nishatoma:
Credits: @evgenru for the idea and good reproducer.

Fixes #1486 and #1420.

Allow consumers to clear `stubRepository` with a public method.

I thought of adding another boolean flag to `clearAllMocks()` initially but that one is used for both JUnit4 + 5 extension, so didn't want to introduce more complexities. So I ended up just adding a new method instead so as to not touch any existing logic. @evgenru @Raibaz @oleksiyp let me know if you have any thoughts on this.

### DIFF
--- a/modules/mockk/src/commonTest/kotlin/io/mockk/impl/stub/StubRepositoryTest.kt
+++ b/modules/mockk/src/commonTest/kotlin/io/mockk/impl/stub/StubRepositoryTest.kt
@@ -14,8 +14,7 @@ import org.junit.jupiter.api.TestInstance
 import kotlin.test.assertFailsWith
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
-class MockKStubMemoryTest {
-
+class StubRepositoryTest {
     interface Service {
         fun compute(x: Int): Int
     }


Raibaz:
I like this, the change LGTM, thanks!

Would you please be able to add documentation about it in README.md? 🙏

nishatoma:
@Raibaz Thanks for the feedback 🙏

The api table was updated as part of initial commit. Is there any specific thing you would like me to mention in it?

Raibaz:
Ah, you're right, I missed that in my latest review. Sorry 🙏

PR #1531, opened 2026-05-29
fix: incidental reflection of new platform apis crashes tests

jzbrooks:
Implements the proposed solution in #1518 (by @DOssaA) and adds a few tests.

The tests are maybe a novel approach for this codebase, but there's precedent for these types of reflection hardening tests in other projects like junit (https://github.com/junit-team/junit-framework/tree/main/platform-tests/src/test/resources).

Fixes #1518

> [!Note]
> I used an llm to generate the test cases after figuring out how other projects tend to approach reflection-dependent tests.


modules/mockk-agent-android/src/main/kotlin/io/mockk/proxy/android/advice/Advice.kt
```diff
@@ -44,8 +44,6 @@ internal class Advice(
             } else {
                 methodDesc.method
             }
-        } catch (_: NoClassDefFoundError) {
-            null
```
jzbrooks: This is a subtype of `LinkageError`
https://docs.oracle.com/javase/8/docs/api/java/lang/LinkageError.html

modules/mockk-agent-android/src/test/kotlin/io/mockk/proxy/android/advice/AdviceOverrideDetectionTest.kt
```diff
@@ -0,0 +1,130 @@
...
+    fun declaredMethods_onHostWithTrap_failsWhenMissingTypeNotOnClasspath() {
+        val hostClass = loadHostWithTrapClass()
+        val error = assertThrows<NoClassDefFoundError> { hostClass.declaredMethods }
+        assertTrue(
+            error.message?.contains("MissingType") == true ||
+                error.cause?.message?.contains("MissingType") == true,
+        ) {
+            "expected MissingType in NoClassDefFoundError, got: ${error.message}"
+        }
+    }
```
jzbrooks, line 71: Running these tests on older JVMs results in false positives wherein reflection rules were a little more relaxed.

The code under test still behaves properly on those older JVMs because no exception is thrown for missing types adjacent to the method being looked up reflectively.

Issue #1224, opened 2024-02-29
Bug: verify on anyConstructed<Mock>() throw an exception


Gosunet:
## Please remove sections wisely

Hello, i encountered a bug doing this PR https://github.com/mockk/mockk/pull/1174 and following this discussion https://github.com/mockk/mockk/pull/1174#issuecomment-1969343409

When using `verify` on `anyConstructed<MockCls>()` we get an error

```
io.mockk.MockKException: Missing calls inside verify { ... } block.
```

### Expected Behavior

```
verify(exactly = 0) { anyConstructed<MockCls>() }
```
should be equivalent to (and not throw exception) 
```
verify { anyConstructed<MockCls>() wasNot Called } 
```

### Current Behavior

```
verify(exactly = 0) { anyConstructed<MockCls>() }
```
throw an exception
```
io.mockk.MockKException: Missing calls inside verify { ... } block.
```

### Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

#### Steps to Reproduce

To reproduce the bug you can replace in the `ConstructorMockTest.clear`method : 

```
verify { anyConstructed<MockCls>() wasNot Called } 
```
by 
```
verify(exactly = 0) { anyConstructed<MockCls>() }
```

#### Stack trace

```
// -----------------------[ YOUR STACK STARTS HERE ] -----------------------
io.mockk.MockKException: Missing calls inside verify { ... } block.
	at io.mockk.impl.recording.states.VerifyingState.checkMissingCalls(VerifyingState.kt:52)
	at io.mockk.impl.recording.states.VerifyingState.recordingDone(VerifyingState.kt:21)
	at io.mockk.impl.recording.CommonCallRecorder.done(CommonCallRecorder.kt:47)
	at io.mockk.impl.eval.RecordedBlockEvaluator.record(RecordedBlockEvaluator.kt:63)
	at io.mockk.impl.eval.VerifyBlockEvaluator.verify(VerifyBlockEvaluator.kt:30)
	at io.mockk.MockKDsl.internalVerify(API.kt:120)
	at io.mockk.MockKKt.verify(MockK.kt:218)
	at io.mockk.MockKKt.verify$default(MockK.kt:209)
	at io.mockk.it.ConstructorMockTest.clear(ConstructorMockTest.kt:100)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
[...]

PR #1548, opened 2026-08-02
fix: clear every mock kind in a single stub repository pass (#1546)

### LINKED ISSUE #1546: clearAllMocks() traverses the whole stub repository 4 times
MockKDsl.internalClearAllMocks dispatches to four clearAll implementations, each iterating stubRepository.allStubs:

```
if (regularMocks)     implementation.clearer.clearAll(options, currentThreadOnly)
if (objectMocks)      implementation.objectMockFactory.clearAll(options, currentThreadOnly)
if (staticMocks)      implementation.staticMockFactory.clearAll(options, currentThreadOnly)
if (constructorMocks) implementation.constructorMockFactory.clearAll(options, currentThreadOnly)
```

allStubs rebuilds a list on each access (stubs.values.mapNotNull { it.value as? Stub }), so one clearAllMocks() costs 4 × O(all live mocks) to clear each mock once.

In releases up to 1.14.11, CommonClearer, JvmObjectMockFactory and JvmStaticMockFactory are byte-identical unfiltered loops clearing every stub (and the constructor factory delegates to the first), so objectMocks/staticMocks/constructorMocks don't scope anything — each flag just triggers another full clear. master filters those three on stub.mockType, fixing the semantics, but the four traversals remain.

**Expected**: each factory iterates its own registry (or one pass routes stubs by type), so the call is a single traversal.

**Impact**: MockKExtension.finish() runs clearAllMocks() after every test method, and in a long-lived test JVM the stub set keeps growing (e.g. mocks in cached Spring test contexts stay strongly reachable). A JFR profile of one of our test targets put 34% of CPU (789/2335 samples) in MockKExtension.finish, hot frames JvmConstructorMockFactory.clearAll, MockKStub.clear, StubRepository.getAllStubs.

renechoi:
Fixes #1546.

## Summary
- `MockKDsl.internalClearAllMocks` dispatched to `Clearer.clearAll`, `ObjectMockFactory.clearAll`, `StaticMockFactory.clearAll` and `ConstructorMockFactory.clearAll` in turn. All four read `StubRepository.allStubs`, and that property rebuilds a list of every live stub on each access, so a single `clearAllMocks()` walked all live mocks four times to clear each mock once.
- Added a `Clearer.clearAll` overload that takes the four kind flags. `CommonClearer` implements it as one pass that routes each stub to the requested kind, and `internalClearAllMocks` now makes a single call.
- The kind classification moved into an internal `Stub.clearKind`. `CommonClearer` and the three JVM factories each carried their own copy of the same `mockType` checks; now they share one.
- Behaviour is unchanged. The two-argument `Clearer.clearAll` delegates to the new overload with only `regularMocks` set, so it still clears regular and spy mocks only, and the per-kind `clearAll` of the three factories keeps working for callers that drive the factories directly. The flag isolation from #1533 is still covered by `ClearAllMocksFlagIsolationTest`, which passes unmodified.
- `apiDump` adds one line per module for the new overload.

## Measurement
Counting reads of `StubRepository.allStubs` during one `clearAllMocks()` with a regular mock, a spy, an object mock, a static mock and a constructor mock live. Measured with a temporary counter on the property, which is not part of this PR:

| | reads of `allStubs` |
| --- | --- |
| master | 4 |
| this branch | 1 |

Wall clock for one `clearAllMocks()` with 5000 live stubbed mocks, JDK 21 on an M-series Mac, 30 warmup calls then 25 samples. Single machine, single run, so treat the ratio rather than the absolute numbers as the result:

| | min | median |
| --- | --- | --- |
| master | 2409 us | 4167 us |
| this branch | 1330 us | 1589 us |

It is not a 4x win because `Stub.clear` runs once per mock either way. Only the repeated traversal and list rebuild go away.

`CommonClearerTest` pins the single read so the extra passes cannot come back unnoticed. Reintroducing one extra `stubRepository.allStubs` read fails `clearing every kind of mock reads the stub repository once` while the second test stays green.

## Verification
- `./gradlew check` on JDK 21: 842 tests, 0 failures, 19 skipped. All skips are pre-existing `@Ignore`, for example `ConstructorMockTest.clear` for #1224.
- `./gradlew spotlessCheck` and `./gradlew apiDump` are clean.

### DIFF
--- a/modules/mockk-dsl/api/mockk-dsl.api
+++ b/modules/mockk-dsl/api/mockk-dsl.api
@@ -730,6 +730,7 @@ public final class io/mockk/MockKGateway$ClearOptions {
 public abstract interface class io/mockk/MockKGateway$Clearer {
 	public abstract fun clear ([Ljava/lang/Object;Lio/mockk/MockKGateway$ClearOptions;)V
 	public abstract fun clearAll (Lio/mockk/MockKGateway$ClearOptions;Z)V
+	public abstract fun clearAll (Lio/mockk/MockKGateway$ClearOptions;ZZZZZ)V
 	public abstract fun clearAllStubsFromMemory (ZLjava/util/List;)V
 }
 
--- a/modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
+++ b/modules/mockk-dsl/src/commonMain/kotlin/io/mockk/API.kt
@@ -651,18 +651,14 @@ object MockKDsl {
             )
         val implementation = MockKGateway.implementation()
 
-        if (regularMocks) {
-            implementation.clearer.clearAll(options, currentThreadOnly)
-        }
-        if (objectMocks) {
-            implementation.objectMockFactory.clearAll(options, currentThreadOnly)
-        }
-        if (staticMocks) {
-            implementation.staticMockFactory.clearAll(options, currentThreadOnly)
-        }
-        if (constructorMocks) {
-            implementation.constructorMockFactory.clearAll(options, currentThreadOnly)
-        }
+        implementation.clearer.clearAll(
+            options,
+            currentThreadOnly,
+            regularMocks,
+            objectMocks,
+            staticMocks,
+            constructorMocks,
+        )
     }
 
     inline fun internalClearMemory(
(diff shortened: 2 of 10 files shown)


Raibaz:
Nice, thanks for looking into this!

